PR #13, opened 2020-02-13
Update to latest KFP SDK release (0.2.2)

ckadner:
Also adding more detailed setup instructions using a separate build directory for the KFP compiler files to not override the `$GOPATH/../kubeflow/pipelines/sdk`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfp-tekton/13)

### DIFF
--- a/sdk/python/kfp/compiler/compiler.py
+++ b/sdk/python/kfp/compiler/compiler.py
@@ -11,12 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import json
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


sdk/python/kfp/compiler/compiler.py
```diff
@@ -30,7 +30,7 @@
 
 from ._default_transformers import add_pod_env
 
-from ..components._structures import InputSpec
+from ..components.structures import InputSpec
```
Tomcli: I remember all the module under components has underscore prefix. Did they changed recently?

ckadner: yes, see https://github.com/kubeflow/pipelines/commit/c83aff2738f77778802fa9fb034bcacc233d125e

PR #16, opened 2020-02-27
Refactor code to extend KFP SDK compiler

ckadner:
This code change removes the requirement to clone and copy the `kubeflow/pipelines` repository. The Python SDK part of the project has been refactored to only contain the minimal set of classes and methods that needed to be modified in order to generate Tekton YAML instead of Argo YAML. The overriden methods are dynamically linked at runtime by _monkey-patching_ the KFP SDK module.

Additionally adding a simple test case and `setup.py` install script that creates the `dsl-compile-tekton` command line executable which now works independent of the KFP
compiler executable as opposed to replacing it.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfp-tekton/16)


sdk/python/requirements.txt
```diff
@@ -0,0 +1 @@
+kfp
```
Tomcli: Do we want to have a minimum kfp version requirement?

ckadner: We could. I have not identified a technical minimum, but we could use the latest tested version which is 0.2.4

Tomcli: sure, I was having 0.1.x version and some library path has changed since then.

ckadner: I set it to `>=0.2.2` which is the version tested earlier

sdk/README.md
```diff
@@ -1,9 +1,11 @@
 # Compiler for Tekton
 
-There is an [SDK](https://www.kubeflow.org/docs/pipelines/sdk/sdk-overview/) for `Kubeflow Pipeline` for end users to
-define pipelines for AI and ML. The output of the KFP SDK compiler is YAML for [Argo](https://github.com/argoproj/argo).
+There is an [SDK](https://www.kubeflow.org/docs/pipelines/sdk/sdk-overview/) 
+for `Kubeflow Pipeline` for end users to define pipelines for AI and ML.
```
animeshsingh:
```suggestion
for `Kubeflow Pipeline` for end users to define end to end machine learning and data pipelines
```

sdk/README.md
```diff
@@ -1,9 +1,11 @@
 # Compiler for Tekton
 
-There is an [SDK](https://www.kubeflow.org/docs/pipelines/sdk/sdk-overview/) for `Kubeflow Pipeline` for end users to
-define pipelines for AI and ML. The output of the KFP SDK compiler is YAML for [Argo](https://github.com/argoproj/argo).
+There is an [SDK](https://www.kubeflow.org/docs/pipelines/sdk/sdk-overview/) 
+for `Kubeflow Pipeline` for end users to define pipelines for AI and ML. 
+The output of the KFP SDK compiler is YAML for [Argo](https://github.com/argoproj/argo).
 
-Here we update the `Compiler` of the KFP SDK to generate `Tekton` YAML for a basic sequential pipeline. 
+Here we update the `Compiler` of the KFP SDK to generate `Tekton` YAML for
```
animeshsingh:
```suggestion
Please follow the steps outlined below to update the `Compiler` of the KFP SDK with a patch to generate `Tekton` YAML for 
```

sdk/python/build.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+
+# The scripts creates the KFP-to-Tekton compiler package.
+#
+# Usage:
+#   ./build.sh [output_file]
+
+
+target_archive_file=${1:-kfptekton.tar.gz}
```
animeshsingh:
```suggestion
target_archive_file=${1:-kfptekton-sequential.tar.gz}
```

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # workflow template
+        container = convert_k8s_obj_to_json(
+            processed_op.container
+        )
+
+        step = {'name': processed_op.name}
+        step.update(container)
+
+        # template.update(container)
```
animeshsingh: I would add an explanation on top that these next set of functionalities are commented out, and we intend to implement them as the project evolves

sdk/python/tests/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020 kubeflow.org
```
animeshsingh: are these files needed if we don't have any code in them?

sdk/python/tests/compiler/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020 kubeflow.org
```
animeshsingh: same as above?

sdk/python/tests/compiler/testdata/sequential.py
```diff
@@ -0,0 +1,51 @@
...
+# limitations under the License.
+
+import kfp
+from kfp import dsl
+
+
+def gcs_download_op(url):
+    return dsl.ContainerOp(
+        name='GCS - Download',
+        image='google/cloud-sdk:216.0.0',
```
animeshsingh: what's the gcs sdk used for?

sdk/python/build.sh
ckadner: @animeshsingh -- this file is the Python install package (`pip install kfptekton.tar.gz`) for the KFP-Tekton SDK/compiler. This is not the (default) filename for a sequential pipeline. 

If you want to the installable filename to reflect the current limitation of only supporting sequential pipeline compilation, then we would have to change it again after my next change to enable parallel pipelines.

sdk/python/kfp_tekton/compiler/_op_to_template.py
ckadner: That would be a short-lived comment as I am changing them currently for my next PR, generating `tasks` instead of `steps`

sdk/python/tests/__init__.py
ckadner: yes, Python uses the mere existing of the `__init__.py` files to indicate a folder being a Python package

sdk/python/tests/compiler/__init__.py
ckadner: same as above :-)

sdk/python/tests/compiler/testdata/sequential.py
ckadner: it's just the example chosen by @vincent-pli using the GCS SDK to download a file from Google Cloud Storage

sdk/python/kfp_tekton/compiler/_op_to_template.py
ckadner: added

PR #40, opened 2020-03-19
Add volumemount support and fix parameter bugs

### LINKED ISSUE #34: add support for volumemounts
/assign @Tomcli

Tomcli:
Preview for volumemount support. Tekton needs to pass param under path `/tekton/results/`. Pending with PR https://github.com/kubeflow/pipelines/pull/3316 to get the config path support on the watson example.

Fixes #34

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfp-tekton/40)


sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -115,11 +115,11 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
       for tp in task.get('params', []):
         if tp['name'] in pipeline_param_names:
           tp['value'] = '$(params.%s)' % tp['name']
-          break
```
ckadner: good catch!

sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -48,6 +48,13 @@ def test_parallel_join_workflow(self):
     from .testdata.parallel_join import download_and_join
     self._test_pipeline_workflow(download_and_join, 'parallel_join.yaml')
 
+  def test_waston_ml_workflow(self):
+    """
+    Test compiling a Waston ML workflow.
+    """
+    from .testdata.watson_train_serve_pipeline import kfp_wml_pipeline
+    self._test_pipeline_workflow(kfp_wml_pipeline, 'watson_train_serve_pipeline.yaml')
```
ckadner: The Watson ML pipeline is a good example of a complex pipeline, but for our automated unit tests we should have a slightly less complex DSL script. The point of unit tests is to tease out individual functionalities, like volume mounts here, but the Watson ML pipeline also uses `secrets`, `image_pull_policy` and requires a GitHub Token to run, which will be the next step with the upcoming integration tests.

sdk/python/tests/compiler/testdata/watson_train_serve_pipeline.py
```diff
@@ -0,0 +1,103 @@
...
+
+# generate default secret name
+import os
+import kfp
+from kfp import components
+from kfp import dsl
+
+secret_name = 'kfp-creds'
+configuration_op = components.load_component_from_url('https://raw.githubusercontent.com/kubeflow/pipelines/master/components/ibm-components/commons/config/component.yaml')
+train_op = components.load_component_from_url('https://raw.githubusercontent.com/tomcli/pipelines/master/components/ibm-components/watson/train/component.yaml')
```
ckadner: `tomcli/pipelines` ... can this point to a public repo? (same below)

sdk/README.md
```diff
@@ -41,7 +47,8 @@ Here we update the `Compiler` of the KFP SDK to generate `Tekton` YAML for a bas
 
 5. Compile the sample pipeline:
 
-    - `cd sdk/samples`  
+    - `mkdir temp && cd temp`
```
ckadner: Linking to external pipeline is okay for complex samples but for basic examples we will need local files in the `sdk/python/tests/compiler/testdata` folder along with the golden output YAML files. I don't think we should get rid of those. So for sequential and parallel join we could point to our local `testdata` folder

sdk/python/tests/compiler/compiler_tests.py
ckadner: the [KFP testdata](https://github.com/kubeflow/pipelines/tree/master/sdk/python/tests/compiler/testdata) folder has:
```
volume.py
volume_snapshotop_rokurl.py
volume_snapshotop_sequential.py
volumeop_basic.py
volumeop_dag.py
volumeop_parallel.py
volumeop_sequential.py
```

sdk/README.md
Tomcli: Hi Christian, I do plan to keep the test files under `sdk/python/tests/compiler/testdata`, but I want to remove `sdk/samples` since it's duplicating with what we have in unittest. For the basic dry run, you think is better to use the unittest files instead of pulling from kfp?

sdk/python/tests/compiler/testdata/watson_train_serve_pipeline.py
Tomcli: Thanks, this is pending on another pr that has the necessary changes. I will update this once we get more information tomorrow.

sdk/python/tests/compiler/compiler_tests.py
Tomcli: Thanks, I will update the unittest.

sdk/python/tests/compiler/testdata/watson_train_serve_pipeline.py
ckadner: ah, right, I should have connected the dots here ...you mentioned it in your opening PR comment  😊

sdk/README.md
ckadner: I want the instructions on the README to be as short and convenient as possible, trying to avoid any extra `mkdir`, `cd`, `curl` commands.

But I understand the value of not keeping copies of files from other repositories, so maybe we can evolve one good example that is not a copy from another repository. Ideally one that shows some extra value from using Tekton over Argo, but that may be a discussion for another issue :-)

sdk/python/tests/compiler/compiler_tests.py
Tomcli: i updated the unit test to use `volume.py` since snapshotop and volumeop is not yet supported.

Tomcli:
The volumes support (not ops) itself is not pending the pr, but in order to run the watson ML with our current implementation, kubeflow/pipelines#3316 is necessary.

sdk/python/tests/compiler/testdata/volume.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2018 Google LLC
```
ckadner: `# Copyright 2020 kubeflow.org` ?

Tomcli: done

sdk/python/tests/compiler/testdata/volume.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Google LLC
+# Copyright 2020 Google LLC
```
ckadner: should this also say `kubeflow.org` instead of `Google LLC` ? (cc @animeshsingh)

Tomcli: Oh i was coping from the kubeflow example, I can update the header

animeshsingh: yeah all licenses should be kubeflow.org

PR #43, opened 2020-03-20
Use apps/v1 instead of extensions/v1beta1

fenglixa:
Change deployment by using apps/v1 instead of extensions/v1beta1, as mentioned here : https://github.com/kubernetes/kubernetes/issues/61430

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfp-tekton/43)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,9 @@ bin
 .venv/
 *.egg-info/
 
+# VSCode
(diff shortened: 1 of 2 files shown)


ckadner:
I think this is good. @Tomcli can you take a look?

PR #50, opened 2020-03-23
Add init_container support

Tomcli:
Adding init_container as initial steps since the execution order is the same. 

Since init_container is not a valid field in Tekton task, I prepend all the init_containers to run them before the main step. Also moving the initContainers before inputs/outputs processing in order to process params for all steps.

### DIFF
--- a/sdk/python/tests/compiler/testdata/init_container.py
+++ b/sdk/python/tests/compiler/testdata/init_container.py
@@ -1,0 +1,29 @@
+# Copyright 2020 kubeflow.org
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


sdk/python/tests/compiler/testdata/init_container.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: tekton.dev/v1beta1
```
fenglixa: add the license header

sdk/python/tests/compiler/testdata/init_container.py
```diff
@@ -0,0 +1,29 @@
...
+    image='alpine:latest',
+    command=['echo', 'bye'])
+
+@dsl.pipeline(name='InitContainer', description='A pipeline with init container.')
+def init_container_pipeline():
+    dsl.ContainerOp(
+        name='hello',
+        image='alpine:latest',
+        command=['echo', 'hello'],
+        init_containers=[echo])
```
fenglixa: Add newline at end of file

sdk/python/tests/compiler/testdata/init_container.yaml
```diff
@@ -0,0 +1,31 @@
...
+    pipelines.kubeflow.org/pipeline_spec: '{"description": "A pipeline with init container.",
+      "name": "InitContainer"}'
+  name: initcontainer
+spec:
+  params: []
+  tasks:
+  - name: hello
+    params: []
+    taskRef:
+      name: hello
```
fenglixa: Add newline at end of file

PR #62, opened 2020-03-25
Add initial pipelinerun support

Tomcli:
Add initial pipelinerun support for development purpose. This PR didn't add any breaking change to the existing compiler. Features such as pipeline timeout and workspaces will come in the following PR since those are breaking features that might require monkey patches on the pipeline dsl definition.

Example usage:
```bash
dsl-compile-tekton --py sdk/python/tests/compiler/testdata/parallel_join.py --output pipeline.yaml --generate-pipelinerun
```
or in Python
```python
TektonCompiler().compile(download_and_join, __file__.replace('.py', '.yaml'), generate_pipelinerun=True)
```

I will also extend it to the developer guide once it's ready.


ckadner:
Thanks @Tomcli, I will take a look.

ckadner:
Hi @Tomcli, 

sorry for the delay. I am still thinking about this. I don't like dragging in ever more of the KFP compiler code, only to accommodate small bits of code changes. But due to the sub-optimal code structure small changes sometimes come with a long tail.

Besides that we have an unfortunate mix of inheritance with overriding classes and class methods, but we rely on monkey-patching to dynamically replace static methods. So we need to ensure all possible entry points to the compiler, like main, CLI, unit tests, direct instantiations all trigger the monkey-patch but not when the actual KFP functions are desired (i.e. using `dsl-compile` on a Python script that has a `__main__` using the `TektonCompiler`).

And conceptually I prefer generating `Pipelines` over `PipelineRuns`, so maybe I just have to get over myself and become a little less opinionated 😉 

I will take a little break and get back to this with fresh eyes a little later.

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -46,6 +46,18 @@ def my_pipeline(a: int = 1, b: str = "default value"):
   ```
   """
 
+  def compile(self, pipeline_func, package_path, type_check=True, pipeline_conf: dsl.PipelineConf = None, generate_pipelinerun=False):
```
ckadner: can you move the `compile` function down to just above `_write_workflow` so that when we compare code changes with KFP all the methods line up

and break the line before the `generate_pipelinerun` argument to keep the line length below 120 for code style compliance and better readability

Tomcli: Thanks @ckadner, I moved compile function.

PR #65, opened 2020-03-26
Add more k8s resource support

Tomcli:
Add Affinity, Node Selector, and Tolerations support with pipelinerun.podTemplate. There are some limitations with this approach where this podTemplate are applied to all tasks in the same pipeline. Also, we only can have one pipelinerun.podTemplate, which means there's only one form of Affinity, Node Selector, and Tolerations can be applied per pipeline. 

@afrittoli @fenglixa do you know is there any way to apply different podTemplate on each individual task basis in a pipelinerun?


fenglixa:
@Tomcli podTemplate for each tasks could be defined in pipeline section, which is same as taskrun. 
Refer to [here](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#pod-template)
> Specifies a pod template configuration that will be used as the basis for the Task pod. This allows to 
> customize some Pod specific field per Task execution, aka TaskRun

sdk/python/tests/compiler/testdata/tolerations.py
```diff
@@ -0,0 +1,35 @@
...
+    op1 = dsl.ContainerOp(
+        name='download',
+        image='busybox',
+        command=['sh', '-c'],
+        arguments=['sleep 10; wget localhost:5678 -O /tmp/results.txt'],
+        file_outputs={'downloaded': '/tmp/results.txt'})\
+        .add_toleration(V1Toleration(effect='NoSchedule',
+                                     key='gpu',
+                                     operator='Equal',
+                                     value='run'))
```
fenglixa: if __name__ == '__main__':
    # don't use top-level import of TektonCompiler to prevent monkey-patching KFP compiler when using KFP's dsl-compile
    TektonCompiler().compile(tolerations, __file__.replace('.py', '.yaml'))

fenglixa:
```suggestion
                                     value='run'))

if __name__ == '__main__':
    # don't use top-level import of TektonCompiler to prevent monkey-patching KFP compiler when using KFP's dsl-compile
    from kfp_tekton.compiler import TektonCompiler
    TektonCompiler().compile(tolerations, __file__.replace('.py', '.yaml'))

```

Tomcli:
> @Tomcli podTemplate for each tasks could be defined in pipeline section, which is same as taskrun.
> Refer to [here](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#pod-template)
> 
> > Specifies a pod template configuration that will be used as the basis for the Task pod. This allows to
> > customize some Pod specific field per Task execution, aka TaskRun

@fenglixa I don't see a podTemplate field in the [pipeline section](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#pipelines). Do you know how to include pod specific field per Task execution in pipeline or pipelinerun? Thanks.

I tried with the below yaml but seems like pipeline api doesn't have any podTemplate field.
```
kind: Pipeline
metadata:
  annotations:
    pipelines.kubeflow.org/pipeline_spec: '{"description": "A pipeline with tolerations",
      "name": "tolerations"}'
  name: tolerations
spec:
  params: []
  tasks:
  - name: download
    params: []
    taskRef:
      name: download
    podtemplate:
      tolerations:
      - effect: NoSchedule
        key: gpu
        operator: Equal
        value: run
```
```
Error from server (BadRequest): error when creating "sdk/python/tests/compiler/testdata/tolerations.yaml": admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "podtemplate"
```

sdk/python/tests/compiler/testdata/tolerations.py
Tomcli: done

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -136,7 +137,7 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
         task['timeout'] = '%ds' % op.timeout
 
     # generate the Tekton Pipeline document
-    pipeline = {
+    pipeline_template = {
```
ckadner: 👍

fenglixa:
Yes, I tried the cases in my test env just now, seems pod template could be only specified for TaskRun or PipelineRun.

sdk/README.md
```diff
@@ -81,3 +81,16 @@ Here we update the `Compiler` of the KFP SDK to generate `Tekton` YAML for a bas
 ## Test Kubeflow Pipelines with Tekton
 
 Please [refer to the instructions here](./python/tests/README.md) as you work on a PR test sample Kubeflow Pipelines in their test data folder to ensure your PR is improving the number of successful samples
+
+## Compile Kubeflow Pipelines as Tekton pipelineRun
+
+Compiling Kubeflow Pipelines into Tekton pipelineRun is currently under the experimental stage. Below are the features that only supported within pipelineRun
```
animeshsingh: This is a good place to add couple of details and difference between pipelinRun and static pipeline, and why pipelineRun is closer to Argo yaml

sdk/README.md
```diff
@@ -81,3 +81,16 @@ Here we update the `Compiler` of the KFP SDK to generate `Tekton` YAML for a bas
...
+- Affinity
+- Node Selector
+- Tolerations
```
animeshsingh: How many total features are there, can we add here?

Tomcli: Right now these are the only three features supported in pipelineRun. There will be more coming in the future RPs.

animeshsingh: yeah - but how many total are there?

Tomcli: I added a link pointing to the list of pipelinerun features.

sdk/README.md
Tomcli: I added reasons why we need to generate pipelinerun and pointed them to the tekton vs argo section in the design docs.

PR #68, opened 2020-03-27
Extend compiler to support workflow level timeout

fenglixa:
Specifies timeout after which the PipelineRun will fail. If the value of timeout is empty, the default timeout will be applied. If the value is set to 0, there is no timeout.
as below, it's the result of:  task-1 was set timout as 20s, pipelinerun timeout was set as 40s.
```
# tkn pr describe pipeline-includes-two-steps-which-fail-randomly-run
Name:           pipeline-includes-two-steps-which-fail-randomly-run
Namespace:      default
Pipeline Ref:   pipeline-includes-two-steps-which-fail-randomly

🌡️  Status

STARTED          DURATION     STATUS
42 seconds ago   40 seconds   Failed

💌 Message

TaskRun pipeline-includes-two-steps-which-fail-randomly-run-rando-8r5ck has failed (TaskRun &#34;pipeline-includes-two-steps-which-fail-randomly-run-rando-8r5ck&#34; failed to finish within &#34;20s&#34;)

📦 Resources

 No resources

⚓ Params

 No params

🗂  Taskruns

 NAME                                                                TASK NAME          STARTED          DURATION     STATUS
 ∙ pipeline-includes-two-steps-which-fail-randomly-run-rando-8r5ck   random-failure     42 seconds ago   20 seconds   Failed(TaskRunTimeout)
 ∙ pipeline-includes-two-steps-which-fail-randomly-run-rando-wzjqz   random-failure-2   42 seconds ago   40 seconds   Failed(TaskRunTimeout)
```

### DIFF
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -129,6 +129,7 @@ class TektonCompiler(Compiler) :
         task['retries'] = op.num_retries
 
     # add timeout params to task_refs, instead of task.
+    pipeline_conf = pipeline.conf
     for task in task_refs:
       op = pipeline.ops.get(task['name'])
       if op.timeout:
@@ -169,6 +170,10 @@ class TektonCompiler(Compiler) :
           }
         }
       }
+
+      # add workflow level timeout to pipeline run
+      if pipeline_conf.timeout:
+        pipelinerun['spec']['timeout'] = '%ds' % pipeline_conf.timeout
 
       workflow = workflow + [pipelinerun]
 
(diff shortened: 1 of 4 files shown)


sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -129,6 +129,7 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
         task['retries'] = op.num_retries
 
     # add timeout params to task_refs, instead of task.
+    pipeline_conf = pipeline.conf
```
ckadner: this seems unneccessary, we can just use `pipeline.conf` later on

Tomcli: I think `pipeline` get overwritten in the next few lines when defining the tekton pipeline. My pr https://github.com/kubeflow/kfp-tekton/pull/65 changes the tekton pipeline variable so it won't conflict with the original pipeline object.

PR #69, opened 2020-03-27
enhance testdata scripts

fenglixa:
Just to enhance deliveried testdata scripts

### DIFF
--- a/sdk/python/tests/compiler/testdata/init_container.py
+++ b/sdk/python/tests/compiler/testdata/init_container.py
@@ -27,3 +27,8 @@ def init_container_pipeline():
         image='alpine:latest',
         command=['echo', 'hello'],
         init_containers=[echo])
+
+if __name__ == '__main__':
+    # don't use top-level import of TektonCompiler to prevent monkey-patching KFP compiler when using KFP's dsl-compile
+    from kfp_tekton.compiler import TektonCompiler
+    TektonCompiler().compile(init_container_pipeline, __file__.replace('.py', '.yaml'))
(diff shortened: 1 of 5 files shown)


fenglixa:
@Tomcli  could you please help reivew? thanks!

Tomcli:
@ckadner I think these pipelines didn't have main function because we want to be aligned with the kfp testdata. Do we want to add main functions to all the test pipelines we have?

ckadner:
I don't think the test data scripts should have a main. They are not intended to be run directly via the Python command line.

I had originally included the main function in the samples to show how to run the `TektonCompiler` and then especially to make it clear that we should not use top-level imports in the scripts, because that triggers the monkey-patch. 

But now that we don't have an example or sample anymore, I think we may want to keep the `__main__` in the `testdata` scripts, until we have an example or sample that has this `__main__` snippet with the import hint.

Tomcli:
Maybe we should pick one or two example from https://github.com/kubeflow/pipelines/tree/master/samples/core as our e2e testing. The ones in testdata are just mock tests for the compiler and not necessary can run successfully on kfp.

ckadner:
> we should pick one or two example from https://github.com/kubeflow/pipelines/tree/master/samples/core as our e2e testing.

Agree. Ideally we would develop our own succinct example, which should demonstrate some of the benefits of Tekton over Argo.

Tomcli:
For now we can merge this in. I will find few examples that could show multiple features as our e2e in some following PR.
/lgtm

ckadner:
> For now we can merge this in. I will find few examples that could show multiple features as our e2e in some following PR.

Yup. I opened issue #71 to track it.

Issue #78, opened 2020-04-01
Map Argo variables to Tekton varaibles.


Tomcli:
KFP uses [Argo variables](https://github.com/argoproj/argo/blob/master/docs/variables.md) to refer artifact locations and exit ops status. Therefore, we want to map as much variables in Argo as possible since issues https://github.com/kubeflow/kfp-tekton/issues/64 and other pipelines assume those variable can be interpreted during execution. 

As of today, Tekton variables $() only has inputs and outputs params on the task level. For other variables, we will try to map them with env variables using Kubernetes FieldRef Parameter. 

However, many Argo variables are still not possible with Tekton variables and pure Kubernetes FieldRef Parameter. Thus, we will raise a warning for users to avoid using them in a pipeline.

Tomcli:
Currently this is the equivalent variable matching i found.
argo -> tekton
`{{inputs.parameters.%s}}` -> `$(inputs.params.%s)`
`{{outputs.parameters.%s}}` -> `$(results.%s.path)`

Since we don't have other equivalent matching for the rest of the [argo variables](https://github.com/argoproj/argo/blob/master/docs/variables.md), we will prompt a warning when those argo variables are used. Since KFP also don't recommend to use Argo variables because it requires specific knowledge on Argo pipeline. 

The other places argo variables are used is during artifact output path. We can workaround with this by creating our own s3/gcs tasks to parse artifact output path using environment variables. This way we don't have to rely on Tekton to replace those workflow variables during execution. However, this means we need to maintain our own s3/gcs task for artifact storing.

ckadner:
Is this what you meant with ...
>  Handler for checking Argo specific params, find equivalent for Tekton

in https://github.com/kubeflow/kfp-tekton/issues/18#issuecomment-600734388 ?

Tomcli:
yes

PR #80, opened 2020-04-02
Add argo variable checking when compile to Tekton.

Tomcli:
Related #78 

Currently this is the equivalent variable matching i found.
argo -> tekton
{{inputs.parameters.%s}} -> $(inputs.params.%s)
{{outputs.parameters.%s}} -> $(results.%s.path)

Since we don't have other equivalent matching for the rest of the argo variables, we will prompt a warning when those argo variables are used. KFP also don't recommend to use [argo variables](https://github.com/argoproj/argo/blob/master/docs/variables.md) because it requires specific knowledge on Argo pipeline.

Out of all the kfp examples, only a very few of them are using Argo variables. e.g. https://github.com/kubeflow/pipelines/blob/master/samples/contrib/nvidia-resnet/pipeline/src/pipeline.py
In this case we will prompt the following warning:
```
Applying KFP-Tekton compiler patch
KFP-Tekton Compiler 0.0.1
WARNING:root:These Argo variables are not supported in Tekton Pipeline: {{workflow.name}}
```

The other places that use Argo variables are during artifact storing and [pipeline placeholder](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/__init__.py#L29-L30). We could create our own tasks to avoid using Argo specific variables on these features as mentioned in #78.

**Environment tested:**

* Kubernetes Version (use `kubectl version`): 1.15

### DIFF
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -197,6 +197,8 @@ class TektonCompiler(Compiler) :
 
     # Use regex to replace all the Argo variables to Tekton variables.
     # For variables that are unique to Argo, we pop a warning to warn the users to avoid critical errors.
+    # Since Argo variables can be used in anywhere of the yaml, we need to dump and parse the whole yaml
+    # using regular expression.
     workflow_dump = json.dumps(workflow)
     tekton_var_regex_rules = [
         {'argo_rule': '{{inputs.parameters.([^ \t\n.:,;{}]+)}}', 'tekton_rule': '$(inputs.params.\g<1>)'},
@@ -209,7 +211,6 @@ class TektonCompiler(Compiler) :
     if unsupported_vars:
       logging.warning('These Argo variables are not supported in Tekton Pipeline: %s' % ", ".join(str(v) for v in set(unsupported_vars)))
     workflow = json.loads(workflow_dump)
-
 
     return workflow  # Tekton change, from return type Dict[Text, Any] to List[Dict[Text, Any]]
 
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -17,7 +17,6 @@ import tarfile
 import yaml
 import zipfile
 import re
-import logging
 from typing import Callable, Set, List, Text, Dict, Tuple, Any, Union, Optional
 
 from ._op_to_template import _op_to_template
@@ -195,8 +194,9 @@ class TektonCompiler(Compiler) :
 
       workflow = workflow + [pipelinerun]
 
-    # Use regex to replace all the Argo variables to Tekton variables.
-    # For variables that are unique to Argo, we prompt a warning to warn users on the non-supported variables.
+    # Use regex to replace all the Argo variables to Tekton variables. For variables that are unique to Argo,
+    # we prompt a warning to warn users on the non-supported variables. Here is the list of Argo variables.
+    # https://github.com/argoproj/argo/blob/master/docs/variables.md
     # Since Argo variables can be used in anywhere in the yaml, we need to dump and then parse the whole yaml
     # using regular expression.
     workflow_dump = json.dumps(workflow)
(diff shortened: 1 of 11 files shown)


sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -191,9 +193,25 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
...
+    tekton_var_regex_rules = [
+        {'argo_rule': '{{inputs.parameters.([^ \t\n.:,;{}]+)}}', 'tekton_rule': '$(inputs.params.\g<1>)'},
+        {'argo_rule': '{{outputs.parameters.([^ \t\n.:,;{}]+).path}}', 'tekton_rule': '$(results.\g<1>.path)'}
+    ]
+    for regex_rule in tekton_var_regex_rules:
+      workflow_dump = re.sub(regex_rule['argo_rule'], regex_rule['tekton_rule'], workflow_dump)
+
+    unsupported_vars = re.findall(r"{{[^ \t\n:,;{}]+}}", workflow_dump)
+    if unsupported_vars:
+      logging.warning('These Argo variables are not supported in Tekton Pipeline: %s' % ", ".join(str(v) for v in set(unsupported_vars)))
```
ckadner, line 212: I am assuming that if one of the Argo specific variables is not supported, then the resulting Tekton pipeline does not behave in the way the author of the Python DSL script intended it to behave. In which case I think should be a little stronger here and `raise TypeError` or `raise ValueError` or even `raise AttributeError`. No?

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -191,9 +193,25 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
       if pipeline_conf.timeout:
         pipelinerun['spec']['timeout'] = '%ds' % pipeline_conf.timeout
 
-
       workflow = workflow + [pipelinerun]
 
+    # Use regex to replace all the Argo variables to Tekton variables.
+    # For variables that are unique to Argo, we prompt a warning to warn users on the non-supported variables.
```
ckadner, line 199: can we list which specific variables we cannot support?

ckadner:
Hi @Tomcli,
do you have a list of all the valriable expressions which cannot be translated to Tekton? 

And if/when you do, are there issues tracking whether those might be in the works by the Tekton team?

Tomcli:
The list is basically this one minus the two I implemented above.
https://github.com/argoproj/argo/blob/master/docs/variables.md

A lot of these variables might now even make sense in Tekton, and most kfp pipelines try to avoid using variables that only Argo could understand.

Tomcli:
Some variables such as task name is already been requested in issue such as https://github.com/tektoncd/pipeline/issues/2294. We can also open another issue on Tekton for this but probably we won't able to get all the variables.

Tomcli:
I opened a related issue at Tekton https://github.com/tektoncd/pipeline/issues/2322

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: Let me see is there other places that rely on this. If not then I can change it to raise and exit.

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: I can add a link to the issue/doc since the variables will change over time.

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -195,8 +194,9 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
 
       workflow = workflow + [pipelinerun]
 
-    # Use regex to replace all the Argo variables to Tekton variables.
-    # For variables that are unique to Argo, we prompt a warning to warn users on the non-supported variables.
+    # Use regex to replace all the Argo variables to Tekton variables. For variables that are unique to Argo,
+    # we prompt a warning to warn users on the non-supported variables. Here is the list of Argo variables.
```
ckadner, line 198: "`# we raise an Error to alert users about the unsupported`" ... :-)

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -209,7 +209,7 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
 
     unsupported_vars = re.findall(r"{{[^ \t\n:,;{}]+}}", workflow_dump)
     if unsupported_vars:
-      logging.warning('These Argo variables are not supported in Tekton Pipeline: %s' % ", ".join(str(v) for v in set(unsupported_vars)))
+      raise ValueError('These Argo variables are not supported in Tekton Pipeline: %s' % ", ".join(str(v) for v in set(unsupported_vars)))
```
ckadner, line 212: 👍

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: thanks christian.

PR #81, opened 2020-04-03
Have Travis/CI enforce that the compilation report is updated

ckadner:
Change the `test_kfp_samples.sh` script such that if returns exit code `1` if the the KFP testdata compilation report changes.

Previously it only checked if the number of `SUCCESS`ful compilations declined from the previous report. This behavior resulted in an outdated compilation report after several PRs missed to update the report.

### DIFF
--- a/sdk/python/tests/README.md
+++ b/sdk/python/tests/README.md
@@ -1,24 +1,31 @@
 ## Tests for the Compiler
 
 
-The test listed here can be used to compile all Python DSL pipelines in the KFP compiler testdata folder and generate a report card. As you are doing a PR to address functionality gaps in compiler, please run this test to ensure that they have been addressed. Please note that even if a Kubeflow Pipeline Python DSL script may pass the compilation with the KFP-Tekton compiler successfully, the produced Tekton YAML might not be valid or contain all of the intended functionality as the equivalent Argo YAML produced by the KFP compiler. For that, best way is to take the compiled YAML and run it on Tekton directly.
+The test listed here can be used to compile all Python DSL pipelines in the KFP compiler `testdata` folder and 
+generate a report card. As you are working a PR to address functionality gaps in the compiler, please run this test to
+update the `FAILURE`s which have been addressed.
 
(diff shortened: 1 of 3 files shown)


fenglixa:
Thanks @ckadner 
/lgtm

ckadner:
In order to resolve the problem with merge conflicts, I will remove the tallies from the bottom of the report file with the number of files which compile successfully or fail. 

The merge conflict which arises when multiple PRs change the report tallies, once one PR gets merged that also changes the same lines. The only way to avoid that would be by removing the totals from the `test_kfp_samples_report.txt` file. The totals will still be printed to the console when running the `test_kfp_samples_report.sh` script. 

I will look into other ways to show the current status, like using Travis/CI to do auto-update     or displaying like a build status on the repo landing page.

ckadner:
@fenglixa @Tomcli -- in order to resolve the problem with merge conflicts, I removed the tallies from the bottom of the report txt file. The totals will still show in the console output. The report txt file will only contain the KFP test file names with the `SUCCESS` or `FAILURE` tag.

fenglixa:
Thanks @ckadner , just one question about this PR: 
Does the file of "test_kfp_samples_report.txt" will be updated automatically? as  you know, if it need to be updated in each PRs, then the test on it will be meanless.

ckadner:
> Thanks @ckadner , just one question about this PR:
> Does the file of "test_kfp_samples_report.txt" will be updated automatically? as you know, if it need to be updated in each PRs, then the test on it will be meanless.

No, we don't yet have automatic update. The Travis job will fail if the `test_kfp_samples_report.txt` does not reflect the current compile status. That means when a PR adds or removes support for certain features, then the `test_kfp_samples_report.txt` need to be updated still. But since the numbers at the bottom are removed now, there should no longer be merge conflicts, unless two PRs make changes on the same feature, which should not happend.

fenglixa:
Thanks @ckadner,  for your clarify, It's OK to me.
/lgtm

PR #84, opened 2020-04-07
Add support for conditions

### LINKED ISSUE #33: Add support for conditions
Add support for condition resources: https://github.com/tektoncd/pipeline/blob/master/docs/conditions.md

A direct example in kfp: https://github.com/kubeflow/pipelines/blob/0.2.2/samples/core/condition/condition.py

drewbutlerbb4:
Related #33

Add support for conditions
Due to Tekton not currently supporting parameter passing from task output to conditions the compiler gracefully exits if it receives a Python DSL with this functionality
**Environment tested:**

**Python 3.7.4**
**Client version: 0.8.0
Pipeline version: v0.11.0-rc2**
**Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:11:03Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8+IKS", GitCommit:"c6910e70b932ed9ffe5baedade2274a72925660f", GitTreeState:"clean", BuildDate:"2020-03-13T14:13:41Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}**
**macOS Mojave 10.14.6**


drewbutlerbb4:
@Tomcli

Tomcli:
Thanks @drewbutlerbb4, can you add a simple test case for what is working right now?

drewbutlerbb4:
A test has been added and I updated test_kfp_samples_report.txt

Note that both coin.py and recursive_do_while.py now fail to compile because they use parameter passing from task outputs to conditions.

Tomcli:
Thanks @drewbutlerbb4, we should not resolves #33 since it's not fully complete. Can you say it is related to #33 from your description and describe what feature we need in Tekton. (e.g. Condition needs to support parameter outputs from the beta API.)

drewbutlerbb4:
Finished both of the above suggestions

sdk/python/tests/test_kfp_samples_report.txt
```diff
@@ -2,15 +2,15 @@ SUCCESS: add_pod_env.py
 SUCCESS: artifact_location.py
 SUCCESS: basic.py
 FAILURE: basic_no_decorator.py
-SUCCESS: coin.py
+FAILURE: coin.py
```
animeshsingh, line 5: So this one is reverting back to failure @drewbutlerbb4 ?

drewbutlerbb4: coin.py uses parameter passing from task outputs to conditions which is [not yet supported in tekton](https://github.com/tektoncd/pipeline/issues/2336). Right now I have the compiler exit gracefully when it sees this so as to not cause confusion.

drewbutlerbb4: This is the same reason recursive_do_while.py fails as well

animeshsingh: Ok - how come it was set to 'SUCCESS' before?

Tomcli: There's a bug in Tekton where we cannot pass output parameters for condition. We get someone in the community to look in this issue https://github.com/tektoncd/pipeline/issues/2336. Once the bug is fixed, then we can implement with output parameters with Tekton 0.12 since Andrew already have that output parameters code for compiler in a different branch.

Tomcli: The compiler only failed if we raise an error. Otherwise it will be ignored and passed.

drewbutlerbb4: It passed before because the compiler ignored conditions altogether. The pipeline would compile fine, but the conditions would not show up in the resulting tekton pipeline yaml.

PR #89, opened 2020-04-09
Add support for dsl.ResourceOp

### LINKED ISSUE #49: Add support for dsl.ResourceOp
The KFP-Tekton compiler does not currently support `dsl.ResourceOp` since Tekton does not provide native support for manipulating Kubernetes resources. 

Argo provides [`ResourceTemplate`](https://github.com/argoproj/argo/blob/53a10564aebc6ee17eb8e3e121b4c36b2a334b87/pkg/apis/workflow/v1alpha1/workflow_types.go#L1395-L1417) to support "resource operations" which the Kubeflow Pipelines SDK wrapped in class [`Resource`/`ResourceOp`](https://github.com/kubeflow/pipelines/blob/1b0c91a091311abe5f43a5d3f6a90d8178516820/sdk/python/kfp/dsl/_resource_op.py#L22)

@afrittoli suggested to implement a standard (set of) Resource Tasks that could be maintained under the [`tektoncd/catalog`](https://github.com/tektoncd/catalog#tekton-catalog) project which provides a library for often reused tasks.

### Proposed implementation

In order to streamline that implementation from a compiler perspective, we could follow the precedence laid out by Argo's `ResourceTemplate`, and create a Tekton Task with the following parameters:

- `action`: `['get', 'create', 'apply', 'delete', 'replace', 'patch']` -- the action to perform to the resource
- `merge_strategy`: `['strategic', 'merge', 'json']` -- the strategy used to merge a patch, defaults to "strategic"
- `success_condition`: a label selector expression which describes the `success` condition
- `failure_condition`: a label selector expression which describes the `failure` condition
- `manifest`: the kubernetes manifest

### Link to issue in `tektoncd/catalog`

[Task(s) to manage kubernetes resources in a cluster #233](https://github.com/tektoncd/catalog/issues/233)

### Additional information

Excerpt from [KFP/Argo-vs-Tekton capabilities spreadsheet](https://docs.google.com/spreadsheets/d/1LFUy86MhVrU2cRhXNsDU-OBzB4BlkT9C0ASD3hoXqpo/edit#gid=979402121&range=22:22)

KFP DSL (kfp.dsl) | Argo Workflow (Current Implementation) | Tekton v0.10.0 (Potential implementation) | Notes | Tekton Backend
-- | -- | -- | -- | --
ResourceOp | ResourceTemplate | Tasks to create, patch, delete, etc. resourcesThe tasks have to be defined, and they could live in the catalog. This is not natively supported by Tekton. | Tekton could define a `ResourceTask`, similar to what Argo does with `ResourceTemplate`, but it is not required to implement ResourceOp. | Workaround

fenglixa:
Resolves #49 

Based on [kubectl-wrapper](https://github.com/vincent-pli/kubectl-wrapper), we have following parameters in Tekton task right now, which are used for supporting Tekton resourceOps:
- **action**:  -- the action to perform to the resource
- **merge_strategy**:  -- the strategy used to merge a patch, defaults to "strategic"
- **success_condition**: a label selector expression which describes the success condition
- **failure_condition**: a label selector expression which describes the failure condition
- **manifest**: the kubernetes manifest
- **output**: An express to retrieval data from resource, all the `output`  will be writen to `tekton/result/output` file.
- **set-ownerreference**: Enable set owner reference for created resource, just leave interfaces there.

@vincent-pli , I think the code of kubectl-wrapper should be committed to kfp-tekton as well, to support more features in future, or any bug fixes.

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_op_to_template.py
+++ b/sdk/python/kfp_tekton/compiler/_op_to_template.py
@@ -16,7 +16,6 @@ from collections import OrderedDict
 from kfp.compiler._k8s_helper import convert_k8s_obj_to_json
 from kfp.compiler._op_to_template import _process_obj, _inputs_to_json, _outputs_to_json
 from kfp import dsl
-from kfp.dsl import ArtifactLocation
 from kfp.dsl._container_op import BaseOp
 
 from .. import tekton_api_version
@@ -106,21 +105,81 @@ def _op_to_template(op: BaseOp):
         }
 
     elif isinstance(op, dsl.ResourceOp):
-        # # no output artifacts
-        # output_artifacts = []
-        #
-        # # workflow template
-        # processed_op.resource["manifest"] = yaml.dump(
-        #     convert_k8s_obj_to_json(processed_op.k8s_resource),
-        #     default_flow_style=False
-        # )
-        # template = {
-        #     'name': processed_op.name,
-        #     'resource': convert_k8s_obj_to_json(
-        #         processed_op.resource
-        #     )
-        # }
-        raise NotImplementedError("dsl.ResourceOp is not yet implemented")
+        # no output artifacts
+        output_artifacts = []
+
+        # task template
+        template = {
+            'apiVersion': tekton_api_version,
+            'kind': 'Task',
+            'metadata': {'name': processed_op.name},
+            'spec': {
+                "params": [
+                    {
+                        "description": "Action on the resource",
+                        "name": "action",
+                        "type": "string"
+                    },
+                    {
+                        "default": "strategic",
+                        "description": "Merge strategy when using action patch",
+                        "name": "merge-strategy",
+                        "type": "string"
+                    },
+                    {
+                        "description": "Content of the resource to deploy",
+                        "name": "manifest",
+                        "type": "string"
+                    },
+                    {
+                        "default": "",
+                        "description": "An express to retrieval data from resource.",
+                        "name": "output",
+                        "type": "string"
+                    },
+                    {
+                        "default": "",
+                        "description": "A label selector express to decide if the action on resource is success.",
+                        "name": "success-condition",
+                        "type": "string"
+                    },
+                    {
+                        "default": "",
+                        "description": "A label selector express to decide if the action on resource is failure.",
+                        "name": "failure-condition",
+                        "type": "string"
+                    },
+                    {
+                        "default": "index.docker.io/fenglixa/kubeclient:v0.0.1",  # Todo: The image need to be replaced, once there are official images from tekton
+                        "description": "Kubectl wrapper image",
+                        "name": "image",
+                        "type": "string"
+                    },
+                    {
+                        "default": "false",
+                        "description": "Enable set owner reference for created resource.",
+                        "name": "set-ownerreference",
+                        "type": "string"
+                    }
+                ],
+                'steps': [
+                    {
+                        "args": [
+                            "--action=$(params.action)",
+                            "--merge-strategy=$(params.merge-strategy)",
+                            "--manifest=$(params.manifest)",
+                            "--output=$(params.output)",
+                            "--success-condition=$(params.success-condition)",
+                            "--failure-condition=$(params.failure-condition)",
+                            "--set-ownerreference=$(params.set-ownerreference)"
+                        ],
+                        "image": "$(params.image)",
+                        "name": processed_op.name,
+                        "resources": {}
+                    }
+                ]
+            }
+        }
 
     # initContainers
     if processed_op.init_containers:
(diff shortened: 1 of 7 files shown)


sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -124,6 +124,14 @@ def test_timeout_workflow(self):
     from .testdata.timeout import timeout_sample_pipeline
     self._test_pipeline_workflow(timeout_sample_pipeline, 'timeout.yaml')
 
+  def test_resourceOp_workflow(self):
+    """
+    Test compiling a resourceOp basic workflow.
+    """
+    from .testdata.resourceop_basic import resourceop_basic
+    nf = lambda f: re.sub("{},{.metadata.name}", "{.metadata.name},{}", f)
```
Tomcli, line 132: Can we document why we need to do this substitution? Also, how can a user know what to define for a normalize function for resourceOp outputs? Also is it possible to have one generic regular expression for doing all these substitution?

sdk/python/tests/test_kfp_samples_report.txt
```diff
@@ -8,20 +8,20 @@ SUCCESS: default_value.py
...
 SUCCESS: sidecar.py
 SUCCESS: timeout.py
 SUCCESS: volume.py
-FAILURE: volume_snapshotop_rokurl.py
-FAILURE: volume_snapshotop_sequential.py
-FAILURE: volumeop_basic.py
-FAILURE: volumeop_dag.py
-FAILURE: volumeop_parallel.py
-FAILURE: volumeop_sequential.py
+SUCCESS: volume_snapshotop_rokurl.py
```
Tomcli, line 19: We probably want to raise errors for volume_snapshotop and volumeop because we don't have any test for these Ops yet and the default volume manifest is generated with some Argo assumption.

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -320,6 +320,37 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
...
+            tp['value'] = action
+          if tp.get('name') == "merge-strategy" and merge_strategy:
+            tp['value'] = merge_strategy
+          if tp.get('name') == "success-condition" and success_condition:
+            tp['value'] = success_condition
+          if tp.get('name') == "failure-condition" and failure_condition:
+            tp['value'] = failure_condition
+          if tp.get('name') == "manifest":
+            manifest = yaml.dump(convert_k8s_obj_to_json(op.k8s_resource), default_flow_style=False)
+            tp['value'] = manifest.replace('{{workflow.name}}', pipeline.name)
```
Tomcli, line 349: At some point we probably need to replace this because `{{workflow.name}}` is unique for every workflow run where `pipeline.name` is static. Can we add a todo comment here? Since Tekton already have pipelinerun name in the task's label, if you somehow can pass pipelinerun name using `fieldRef` then it can also work as well.

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -106,21 +105,81 @@ def _op_to_template(op: BaseOp):
...
+                        "type": "string"
+                    },
+                    {
+                        "default": "",
+                        "description": "A label selector express to decide if the action on resource is failure.",
+                        "name": "failure-condition",
+                        "type": "string"
+                    },
+                    {
+                        "default": "index.docker.io/fenglixa/kubeclient:v0.0.1",
```
Tomcli, line 153: Can also add a todo comment to replace this image once our task is pushed to Tekton catalog?

Tomcli:
We probably need to have some instructions for adding certain rbac permission to this pipeline. I will open an issue and do it in a different PR.

sdk/python/tests/compiler/compiler_tests.py
fenglixa: Add to the document. Thanks @Tomcli

sdk/python/tests/test_kfp_samples_report.txt
fenglixa: Addressed, all the volumeOp are failed now.

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: Addressed. Thanks @Tomcli

sdk/python/kfp_tekton/compiler/compiler.py
fenglixa: Just removed it, since resourceOp don't have {{workflow.name}} defined, it was appeared in VolumeOp.

fenglixa:
Thanks @Tomcli , I addressed you comments, please help review whether it's OK for you now.

fenglixa:
> Isn't the wrapper going in Tekton repo?
I discussed it with @vincent-pli before. It's better the wrapper could be going to Tekton repo, but seems it could be only going as Tekton catalog.
Personaly, we should find a location in kfp-tekton, to let the wrapper code there, convenient for us to maintain and add new features, how do you think? @animeshsingh .

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: Ok then I should put it as a todo for the VolumeOp and volumesnapshotop.

Tomcli:
Thanks @fenglixa, it works well with the basic resourceOp example. However, the `output` only takes one argument where volumeOp and volumeSnapShotOp is expecting 3. Probably we need to enhance the `output` argument to output a list of results instead of a single output file. Probably we need to enhance the wrapper for that feature.

Although I do love to merge what we have now, so we can start work on the new features for volumeOp and volumeSnapShotOp in parallel. 

/lgtm

Tomcli:
@fenglixa So after some discussion with Animesh, we probably should push the warpper code to tektoncd/catalog because kfp-tekton shouldn't maintain any custom Tekton image. Also this way we can get more feedbacks from the tekton community. What do you think?

fenglixa:
>So after some discussion with Animesh, we probably should push the warpper code to tektoncd/catalog because kfp-tekton shouldn't maintain any custom Tekton image. Also this way we can get more feedbacks from the tekton community. What do you think?

It's OK to me. Thanks.

PR #91, opened 2020-04-10
Fix special compilation cases for kfp samples

drewbutlerbb4:
Fixes the testing of [kfp samples](https://github.com/kubeflow/pipelines/tree/master/sdk/python/tests/compiler/testdata) for pipelines that require special compilation instructions. 

Specifically these include;
compose.py
basic_no_decorator.py


sdk/python/tests/test_util.py
```diff
@@ -0,0 +1,78 @@
...
+            'Get Most Frequent Word and Save to GCS',
+            [
+                basic_no_decorator.message_param,
+                basic_no_decorator.output_path_param
+            ])
+        print("SUCCESS: basic_no_decorator.py")
+    except:
+        print("FAILURE: basic_no_decorator.py")
+
+def test_composing_workflow(test_data_dir):
```
Tomcli, line 50: Can we generalize this helper function to take any nested pipeline? We also have few nested pipeline in the sample repository that we want to make it pass. You can add arguments to also take the class name and the pipeline function names.

sdk/python/tests/test_util.py
```diff
@@ -0,0 +1,78 @@
...
+import kfp_tekton.compiler as compiler
+import filecmp
+
+def _get_yaml_from_zip(zip_file):
+    with zipfile.ZipFile(zip_file, 'r') as zip:
+        with open(zip.extract(zip.namelist()[0]), 'r') as yaml_file:
+            return list(yaml.safe_load_all(yaml_file))
+
+
+def test_basic_workflow_without_decorator(test_data_dir):
```
Tomcli, line 32: Can we generalize this helper function to take any pipeline without decorator? You can pass any infomation you need as arguments.

sdk/python/tests/test_kfp_samples.sh
```diff
@@ -73,11 +73,15 @@ rm -f "${COMPILER_OUTPUTS_FILE}"
 # compile each of the Python scripts in the KFP testdata folder
 for f in "${KFP_TESTDATA_DIR}"/*.py; do
   echo -e "\nCompiling ${f##*/}:" >> "${COMPILER_OUTPUTS_FILE}"
-  if dsl-compile-tekton --py "${f}" --output "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" >> "${COMPILER_OUTPUTS_FILE}" 2>&1;
-  then
-    echo "SUCCESS: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
+  if [ ${f##*/} == "compose.py" ] || [ ${f##*/} == "basic_no_decorator.py" ]; then
```
Tomcli, line 76: Can we create a config file for all the special pipelines along with any extra information to compile it?

drewbutlerbb4:
@Tomcli I believe I have addressed your comments. I can't test on either of the linked pipelines because they have functionality that kfp-tekton can't currently compile (exithandler and input artifacts). I will make another pipeline and test on that shortly.

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -382,59 +382,62 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
...
-        workflow = workflow + [service_template]
-        pipelinerun['spec']['serviceAccountName'] = service_template['metadata']['name']
+        if service_template:
+          workflow = workflow + [service_template]
+          pipelinerun['spec']['serviceAccountName'] = service_template['metadata']['name']
 
-      workflow = workflow + [pipelinerun]
+        workflow = workflow + [pipelinerun]
+    except:
+      pass # Intentionally do nothing
```
Tomcli, line 440: Can you add a one line comment for why we need to pass the errors for no pipeline decorator?

sdk/python/tests/test_kfp_samples.sh
```diff
@@ -70,14 +71,22 @@ cp "${COMPILE_REPORT_FILE}" "${COMPILE_REPORT_FILE_OLD}"
...
+# check which pipelines have special configurations
+PIPELINES=$(awk '/pipeline:/{print $NF}' ${CONFIG_FILE})
+
 # compile each of the Python scripts in the KFP testdata folder
 for f in "${KFP_TESTDATA_DIR}"/*.py; do
   echo -e "\nCompiling ${f##*/}:" >> "${COMPILER_OUTPUTS_FILE}"
-  if dsl-compile-tekton --py "${f}" --output "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" >> "${COMPILER_OUTPUTS_FILE}" 2>&1;
-  then
-    echo "SUCCESS: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
+  IS_SPECIAL=$(grep -E ${f##*/} <<< ${PIPELINES})
```
ckadner, line 80: For the `test_kfp_samples.sh` script, why do we not just generate `PipelineRun` for all `testdata` scripts and only change one line:

```Bash
  if dsl-compile-tekton --py "${f}" --output "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" >> "${COMPILER_OUTPUTS_FILE}" 2>&1;
```
to
```Bash
  if dsl-compile-tekton --generate-pipelinerun --py "${f}" --output "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" >> "${COMPILER_OUTPUTS_FILE}" 2>&1;
```
There is no need or benefit to have a mix here, since we will catch the `NotImplementedError`s either way.

sdk/python/tests/test_util.py
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
ckadner, line 1: should we not move these tests into our standard compiler test suite?

sdk/python/tests/test_kfp_samples.sh
drewbutlerbb4: Maybe I do not understand, but wouldn't this still fail to compile basic_no_decorator.py and compose.py?

ckadner: Hi Andrew, yes, I wrote that comment before understanding the full purpose of your `test_util.py`. But let me explain the my bigger point below

ckadner:
Hi Andrew,

I think the code you wrote in `test_util.py` is good but it should really be part of our unit test suite for the compiler. In fact the KFP `testdata` scripts are meant to be used in the compiler unit tests where all the negative cases and special parameters are handled.

The `test_kfp_samples.sh` script was really just intended to show us which features still need to be implemented on the compiler and give us a kind of score card. It was not the intention to do all we can to make all tests succeed just to make sure we know what was not yet working.

We are at a time in the project where we should add all KFP unit tests in our unit test suite and use that to measure our progress.

On another tangent I have been using the `test_kfp_samples.sh` script locally to run the compiler over all DSL samples in the KFP repository to get a better sense of which features are actually used by KFP consumers and then use some bash scripting to dissect the compiler console output to computer a list of all errors. I was debating about making that public but it now is kind of conflicting with your change, so I admit some bias on my part :-)

drewbutlerbb4:
Thanks for the explanation Christian I now understand your point of view better. I agree that this could be placed in the compiler tests along with some new example pipelines and that could help improve our testing.

However I am not entirely sure it shouldn't be included in the `test_kfp_samples.sh` as well. Based on discussions from our meetings, I believe the output of this script (as well as your new improvements) is being used as a metric to judge the progress of this project by people not working day-to-day on the kfp-tekton? Wouldn't it then be confusing to these people to see pipelines failing when they shouldn't be? 

I would like to get @Tomcli opinion on your comments because, admittedly I'm not entirely sure this was the intention when I was assigned this work. That being said I'll defer to @ckadner and @Tomcli on this one.

Tomcli:
@ckadner I agree we can also use test_util.py as part of our unit test suite for the compiler, but we should also leverage it in test_kfp_samples.sh for the metrics because nested pipelines and no decorator pipelines need special ways to compile them. If you compile them using the regular `dsl-compile` or `dsl-compile-tekton`, you will see the below 2 errors
```
   1 ValueError: There are multiple pipelines: ['save_most_frequent_word', 'download_save_most_frequent_word']. Please specify --function.
   1 ValueError: A function with @dsl.pipeline decorator is required in the py file.
```

Therefore, I asked Andrew to take those special compile cases and generalize it. This way we can update the list of nested pipelines and no decorator pipelines in the config.yaml and compile them with the required parameters.

sdk/python/tests/test_kfp_samples.sh
```diff
@@ -70,14 +71,22 @@ cp "${COMPILE_REPORT_FILE}" "${COMPILE_REPORT_FILE_OLD}"
...
+  if [ -z "${IS_SPECIAL}" ]; then
+    if dsl-compile-tekton --py "${f}" --output "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" >> "${COMPILER_OUTPUTS_FILE}" 2>&1;
+    then
+      echo "SUCCESS: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
+    else
+      echo "FAILURE: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
+    fi
   else
-    echo "FAILURE: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
+    python3 -m test_util ${f} ${CONFIG_FILE} | grep -E 'SUCCESS:|FAILURE:'
```
ckadner, line 89: this should move up so we can capture the entire output of the compile run in the `${COMPILER_OUTPUTS_FILE}`, not just the `SUCCESS` or `FAILURE`. The `python3` command should return an exit value of 0 for success and 1 for failure

Maybe we could use a ternary stile approach:
```Bash
    if ([ -z "${IS_SPECIAL}" ] \
        && dsl-compile-tekton --py "${f}" --output "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" \
        || python3 -m test_util ${f} ${CONFIG_FILE} ) >> "${COMPILER_OUTPUTS_FILE}" 2>&1 ;
    then
      echo "SUCCESS: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
    else
      echo "FAILURE: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
    fi
```

...or better.. wrap the compile part into a bash function and inside that function check for `IS_SPECIAL`:
```Bash
function compile_dsl {
  if [ -z "${IS_SPECIAL}" ]; then
    dsl-compile-tekton --py "$1" --output "$2"
  else
    python3 -m test_util $1 ${CONFIG_FILE} # need to produce YAML file output
  fi
} 

# ...

    if compile_dsl "${f}" "${TEKTON_COMPILED_YAML_DIR}/${f##*/}.yaml" >> "${COMPILER_OUTPUTS_FILE}" 2>&1 ;
    then
      echo "SUCCESS: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
    else
      echo "FAILURE: ${f##*/}" | tee -a "${COMPILER_OUTPUTS_FILE}"
    fi
```

sdk/python/tests/test_util.py
ckadner: please move the two functions `test_workflow_without_decorator` and `test_nested_workflow` into the actual compiler unit test and have the `test_util.py` call them and capture the output YAML into a file, like `dsl-compile` does

ckadner:
@drewbutlerbb4 -- I realize I am requesting substantial changes to your PR, some of which are quite aspirational and could be scoped out into a separate PR. Since I am working on another PR in the `test_kfp_samples.sh` script anyway, I could incorporate those changes there and we can merge your PR as is. What do you think?

drewbutlerbb4:
Sure that works for me

PR #93, opened 2020-04-14
Add volumeop and volumesnapshot support

### LINKED ISSUE #51: add support for dsl.volumeOps
Documentation for volumeOps: https://github.com/kubeflow/pipelines/blob/master/samples/core/volume_ops/README.md

volumeOps in kfp currently do the following things:
1. Create the PVC.
2. Mount PVC for every component that uses the volumeOps

In the very low level, it uses the [same resource function](https://github.com/kubeflow/pipelines/blob/master/sdk/python/tests/compiler/testdata/volumeop_sequential.yaml#L14-L30) in Argo for creating the PVC resource. Therefore, we can reuse the resourceOps task for volumeOps for the PVC creation.

For the PVC mount, since we already know the name of the pvc and have volume support. We only need to make the resourceOps task to fail fast if something went wrong. 

Alternatively, instead of using the pvc mount, we can use workspaces instead. However, we need to reimplement our own logic for workspaces and remove the existing volumeOps PVC mount logic. Furthermore, we will need to generate a pipelinerun yaml because the tkn cli generator is not yet supporting workspaces code-gen. The same workspaces discussion also in https://github.com/kubeflow/kfp-tekton/issues/41 for using it to copy files between steps.

The proposed [auto workspaces](https://github.com/tektoncd/pipeline/issues/1986) could also work if it's implemented the way as it promised. However, it still going to take some time for someone to start a PR in Tekton and implement this feature.

### LINKED ISSUE #86: support volumeSnapShotOp
/kind feature

**Description:**
support volumeSnapShotOp

**Additional information:**
This issue requires resourceOp to be implemented in #49

Tomcli:
Resolves #51 and resolves #86 

<del> Add volumeop support with some workaround on the unsupported Argo variables and extend task outputs using an extra step. Ideally we should improve the wrapper with multiple outputs. I will create an issue for the wrapper issue so we can improve it in the future. </del>

Update resourceOp with vincent's latest changes to support volumeop and volumesnapshotop.

**Environment tested:**

* Kubernetes Version (use `kubectl version`): 1.15


Tomcli:
Vincent updated the resourceop task, so i will update the volumeop with the corresponding changes soon.
/hold

Tomcli:
FYI, don't test with the default image yet because there's a small bug in the wrapper code. Will unhold this once the image is updated.

Tomcli:
/unhold
Note the [volume snapshot](https://kubernetes.io/docs/concepts/storage/volume-snapshots/) is only available with feature gate unless you are running k8s 1.17 which is also mentioned in the comments.

sdk/python/tests/test_kfp_samples_report.txt
```diff
@@ -8,20 +8,20 @@ SUCCESS: default_value.py
...
-FAILURE: volumeop_basic.py
-FAILURE: volumeop_dag.py
-FAILURE: volumeop_parallel.py
-FAILURE: volumeop_sequential.py
+SUCCESS: volume_snapshotop_rokurl.py
+SUCCESS: volume_snapshotop_sequential.py
+SUCCESS: volumeop_basic.py
+SUCCESS: volumeop_dag.py
+SUCCESS: volumeop_parallel.py
+SUCCESS: volumeop_sequential.py
```
fenglixa, line 24: Just a question here, does all these volumeop works now?

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -326,8 +327,8 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
       if isinstance(op, dsl.ResourceOp):
         action = op.resource.get('action')
         merge_strategy = op.resource.get('merge_strategy')
-        success_condition = op.resource.get('success_condition')
-        failure_condition = op.resource.get('failure_condition')
+        success_condition = op.resource.get('successCondition')
+        failure_condition = op.resource.get('failureCondition')
```
fenglixa, line 331: Great, thanks.

sdk/python/tests/test_kfp_samples_report.txt
Tomcli: These pipelines should work as long as your default storageclass support readwritemany.

Tomcli: I tested the basic and parallel one on my cluster and they are running as expected.

fenglixa:
Thanks @Tomcli 
/lgtm

Tomcli:
/assign @animeshsingh @ckadner
Do you guys have any question/opinion on this PR?

Issue #63, opened 2020-03-25
Fix big data passing


ckadner:
Re-implement the [`fix_big_data_passing`](https://github.com/kubeflow/kfp-tekton/blob/00504fe8493b1c41423dfd5323bff6e1ccbaae04/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py#L18-L65) function, which converts a workflow where some artifact data is passed as parameters and converts it to a workflow where this data is passed as artifacts.

This method is currently only a No-op pass-through in the KFP-Tekton compiler:
```Python
def fix_big_data_passing(workflow: Dict[Text, Any]) -> Dict[Text, Any]: 
    """
    No-op
    """
    return workflow
```

### Motivation

Due to the convoluted nature of the DSL compiler, the artifact consumption and passing has been implemented on top of that using parameter passing. The artifact data is passed as parameters and the consumer template creates an artifact/file out of that data.

Due to the limitations of Kubernetes (and Argo, Tekton) this scheme cannot pass data larger than few kilobytes preventing any serious use of artifacts.

The `fix_big_data_passing` function rewrites the compiled workflow so that the data consumed as artifact is passed as artifact. It also prunes the unused parameter outputs. This is important since if a big piece of data is ever returned through a file that is also output as parameter, the execution will fail. This makes is possible to pass large amounts of data.

### Implementation in KFP

1. Index the DAGs to understand how data is being passed and which inputs/outputs are connected to each other.
2. Search for direct data consumers in container/resource templates and some DAG task attributes (e.g. conditions and loops) to find out which inputs are directly consumed as parameters/artifacts.
3. Propagate the consumption information upstream to all inputs/outputs all the way up to the data producers.
4. Convert the inputs, outputs and arguments based on how they're consumed downstream.

### KFP implementation

See `kubeflow/pipelines`: [`sdk/python/kfp/compiler/_data_passing_rewriter.py`](https://github.com/kubeflow/pipelines/blob/3df29394c1f70298071fc94a77ecd36e9a32d606/sdk/python/kfp/compiler/_data_passing_rewriter.py#L6)

Tomcli:
This is the original code from kfp
https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/_data_passing_rewriter.py

This big data passing has very complex processing to convert local files into artifact passing methods in a very Argo specific way. We would love to see can we reuse some of these methods so that we don't have to maintain our version of big data passing.

[...]

Issue #96, opened 2020-04-16
Refactor compiler and op to template function.


Tomcli:
/kind feature

**Description:**
[A clear and concise description of what your proposal. What problem does it solve?]
As we introduce more features, there are some task/step template that we have to introduce in order to replicate some actions in the Argo sidecar. So we should refactor the code in a way that any additional template can be added as a plugin function.

**Additional information:**
[Miscellaneous information that will assist in solving the issue.]

PR #104, opened 2020-04-21
Support input artifact

### LINKED ISSUE #64: Support artifact outputs
Support artifact outputs where we should store those artifacts to an object storage.

Required resources:
1. s3/gcs secrets 
2. pipelineresource/resourcetemplate
3. resource mapping in pipelinerun (can be generate with tkn cli)
4. Task and pipeline resource mapping (WIP branch https://github.com/Tomcli/kfp-tekton/tree/artifacts)
5. The kfp dsl also treated the output parameters as output artifacts. Do we want to setup a flag to disable it?

Tomcli:
Resolves #64 

This is a follow up PR for #92 to support input artifacts. In KFP, the input artifacts are passed from the parameter outputs and loaded as a file instead of loading it from S3. The implementation details are located at 
https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/_op_to_template.py#L123-L145
https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/_op_to_template.py#L223-L225

From the source code, you can see there's no S3 involved with passing inputs as artifacts in all cases. The only limitation we have right now is missing #63 for big file compression which none of the example or test is using it.

Put it on hold for now until we merged #92 for the artifact support.
/hold

**Environment tested:**


Tomcli:
/unhold
rebased with the master branch

Tomcli:
This will be the last PR before we do the major refactor for #96

PR #119, opened 2020-04-24
Fix edge cases such as disable istio-inject for Tekton compiler

### LINKED ISSUE #114: Docs for deploying tekton pipeline on kfp user namespaces.
/kind feature

**Description:**

We need step by step docs for users who want to deploy tekton on kfp namespace. (such as workaround on istio-inject without [kfp api](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go#L332-L340))

/assign

**Additional information:**

Tomcli:
Resolves #114 

- Add new annotation to disable istio-inject
- Remove literal string conversion for pyyaml since there could be conflicts with yaml 1.1 and 1.2 on some [edge cases](https://github.com/yaml/pyyaml/issues/98). If we want this feature back then we have to use ruamel.yaml. We are using pyyaml right now because we need to be consistent with the kfp package. 
- Add headers to golden test yaml
- Fix minor syntax bugs

**Environment tested:**

* Tekton Version (use `tkn version`): 0.11.3

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_op_to_template.py
+++ b/sdk/python/kfp_tekton/compiler/_op_to_template.py
@@ -26,18 +26,6 @@ import re
 from .. import tekton_api_version
 
 
-class literal_str(str):
-    """Literal string class for pyyaml
-
-    Literal string class is used for converting string with newline into
-    yaml's literal string format with '|'. In pyyaml, literal string
-    conversion is not natively supported in the default dumper.
-    Therefore, we need to define this class as part of the dumper
-    before compiling it into yaml.
-    """
-    pass
-
-
 def _get_base_step(name: str):
     """Base image step for running bash commands.
 
@@ -329,12 +317,11 @@ def _process_output_artifacts(outputs_dict: Dict[Text, Any],
         mounted_artifact_paths = []
         for artifact in outputs_dict['artifacts']:
             if artifact['name'] in replaced_param_list:
-                print(replaced_param_list)
                 copy_artifacts_step['script'] = copy_artifacts_step['script'] + \
-                    'mc cp $(results.%s.path) storage/%s/runs/$PIPELINERUN/$PODNAME/%s' % (artifact_to_result_mapping[artifact['name']], bucket, artifact['path'].rsplit("/", 1)[1])
+                    'mc cp $(results.%s.path) storage/%s/runs/$PIPELINERUN/$PODNAME/%s\n' % (artifact_to_result_mapping[artifact['name']], bucket, artifact['path'].rsplit("/", 1)[1])
             else:
                 copy_artifacts_step['script'] = copy_artifacts_step['script'] + \
-                    'mc cp %s storage/%s/runs/$PIPELINERUN/$PODNAME/%s' % (artifact['path'], bucket, artifact['path'].rsplit("/", 1)[1])
+                    'mc cp %s storage/%s/runs/$PIPELINERUN/$PODNAME/%s\n' % (artifact['path'], bucket, artifact['path'].rsplit("/", 1)[1])
                 if artifact['path'].rsplit("/", 1)[0] not in mounted_artifact_paths:
                     volume_mount_step_template.append({'name': artifact['name'], 'mountPath': artifact['path'].rsplit("/", 1)[0]})
                     volume_template.append({'name': artifact['name'], 'emptyDir': {}})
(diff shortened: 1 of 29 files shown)


sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -236,6 +237,23 @@ def _test_pipeline_workflow(self,
         compiled = list(yaml.safe_load_all(f))
       if GENERATE_GOLDEN_YAML:
         with open(golden_yaml_file, 'w') as f:
+          f.write(textwrap.dedent("""\
```
ckadner, line 240: can you move this text into a variable at the top of the file

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -36,18 +36,6 @@
 from .. import tekton_api_version
 
 
-def _literal_str_representer(dumper, data):
```
ckadner, line 39: I forgot the history of this, can you quickly explain why we had to add this and why it's no longer needed?

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -445,17 +432,6 @@ def _op_to_template(op: BaseOp, enable_artifacts=False):
     if processed_op.init_containers:
         template['spec']['steps'] = _prepend_steps(processed_op.init_containers, template['spec']['steps'])
 
-    # initial base_step and volume setup
-    base_step = {
-        'image': 'busybox',
-        'name': 'copy-results',
-        'script': '#!/bin/sh\nset -exo pipefail\n'
-    }
-    volume_mount_step_template = []
```
ckadner, line 454: aren't those still being used in the code below?

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -445,17 +432,6 @@ def _op_to_template(op: BaseOp, enable_artifacts=False):
     if processed_op.init_containers:
         template['spec']['steps'] = _prepend_steps(processed_op.init_containers, template['spec']['steps'])
 
-    # initial base_step and volume setup
-    base_step = {
-        'image': 'busybox',
-        'name': 'copy-results',
```
ckadner, line 451: I still see the `copy-results` step in the compiled YAML and in a comment, does that come from somewhere else?

Tomcli: I moved this into a different function, but somehow it get put back after the merge. We no longer need this over here because it is initiated at the beginning of this function.

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: I wanted to add literal style to make the script part looks nicer. But there's some edge cases with pyyaml when using it on integer since it's designed with yaml 1.1

Here's the definition of literal style in yaml
https://yaml.org/spec/1.2/spec.html#id2795688

sdk/python/tests/compiler/compiler_tests.py
Tomcli: done

sdk/python/kfp_tekton/compiler/_op_to_template.py
ckadner: right, thanks :-)

sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -226,18 +226,7 @@ def _test_pipeline_workflow(self,
...
-                                        generate_pipelinerun=generate_pipelinerun,
-                                        enable_artifacts=enable_artifacts)
-      with open(compiled_yaml_file, 'r') as f:
-        f = normalize_compiler_output_function(
-          f.read()) if normalize_compiler_output_function else f
-        compiled = list(yaml.safe_load_all(f))
-      if GENERATE_GOLDEN_YAML:
-        with open(golden_yaml_file, 'w') as f:
-          f.write(textwrap.dedent("""\
+    license_header = textwrap.dedent("""\
```
ckadner, line 229: I meant move `license_header` to the top of the `compiler_tests.py` file since Andrew will have to use that in his PR as well and it blows up the code too much here

Tomcli: oh you mean set it as global variable?

sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -27,6 +27,24 @@
 # in order to generate new "golden" YAML files
 GENERATE_GOLDEN_YAML = False
 
+# License header for Kubeflow project
+LICENSE_HEADER = textwrap.dedent("""\
```
ckadner, line 31: nice, thanks :-)

PR #126, opened 2020-04-29
Use make commands for tests and report

ckadner:
Since @kevinyu98 added a neat new `Makefile` we should _make_ use it in our docs and for CI/CD with Travis 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,27 +5,24 @@ matrix:
       python: "3.6"
       env: TOXENV=py36
       install:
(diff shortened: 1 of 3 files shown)


ckadner:
Thanks @Tomcli -- I just realized `make` won't look for the `Makefile` in parent directories, so I updated the wording in the `sdk/python/tests/README.md`

Tomcli:
I think there's a bug with prow tide because it still thinks this PR is not merge. Let me try to close this can see can we skip that bug.
/close

Tomcli:
Problem solved. Looks like reopen reset the tide stage.

PR #148, opened 2020-05-13
Remove compiler monkey-patch

ckadner:
### TL;DR:

All of the _"patched"_ methods are now exclusively called directly by other code in the `kfp_tekton.compiler`, therefore the _**monkey-patch**_ is no longer needed.

### Motivation:

The _monkey-patching_ was a means to reduce the amount of code we have to duplicate. 

In the beginning of the project we really only needed to modify a few compiler methods and some static helper methods. But the KFP code that was calling those methods did not need to also be overridden. So the _monkey-patching_ was used to "inject" our changed code into the call hierarchies of the KFP compiler code dynamically.

Over time, we pulled in more and more code from KFP, eventually covering all of the invocation paths, i.e. from the CLI as well as direct instantiations of the compiler. Now, all of the static helper methods we have "overridden" are exclusively called by other code we have also "overridden" -- leaving no more calling paths to be _patched_ dynamically.

However it may become necessary again in the future to reintroduce the _monkey patch_ for any methods we need to "override" which are not (only) directly called by other methods we already implemented in the `kfp_tekton` compiler. Therefore I opted to leave some documentation in the developer guide.


Tomcli:
Thanks @ckadner, before I review this, is this PR's main motivation is to make our sdk to purely use python inheritance instead of monkey patching?

ckadner:
> ... is this PR's main motivation is to make our SDK to purely use Python inheritance instead of _monkey patching_?

Yes. The _monkey-patching_ was a means to reduce the amount of code we have to duplicate. 

In the beginning of the project we really only needed to modify a few compiler methods and some static helper methods. But the KFP code that was calling those method did not need to also be overridden. So the _monkey-patching_ was used to "inject" our changed code into the call hierarchies of the KFP compiler code.

Over time, we pulled in more and more code from KFP, eventually covering all of the invocation paths, i.e. from the CLI as well as direct instantiations of the compiler. Now, all of the static helper methods we have "overridden" are exclusively called by other code we have also "overridden" -- leaving no more calling paths to be _patched_ dynamically.

sdk/python/README.md
```diff
@@ -129,6 +129,11 @@ When code changes are required to static helper methods in `kfp.compiler` the "o
 their respective modules in `kfp_tekton.compiler` and added to the _monkey patch_ which dynamically replaces the code in
 the `kfp` at runtime.
 
+**Note**: As of May 2020, the _monkey patch_ was no longer needed and removed since all of the _patched_ methods were
```
Tomcli, line 132: I think we should move the monkey patching section to the bottle and mark it as deprecated. This way the users won't get confused if they just reading from middle of the page.

Tomcli:
Everything else looks good to me.

sdk/python/README.md
ckadner: Okay. I moved the detailed description and code snippets for the _monkey patch_ to the bottom of the developer guide:
https://github.com/ckadner/kfp-tekton/blob/remove_monkey_patch/sdk/python/README.md#adding-new-code

But I do want to leave the main paragraph about it in the middle of the doc for future reference since it is one of two main ways to inject code into the KFP compiler and it may not be obvious to other developers how to implement it, or, that it even is a possibility.

Tomcli: Thanks @ckadner

ckadner:
Thanks @Tomcli

PR #156, opened 2020-05-20
Rename main step name to "main"

### LINKED ISSUE #154: Rename Tekton task's main container as "main"
/kind feature

**Description:**

Since Argo always name the main container as "main", it would be easier for the API to figure out which container to get the logs if we can change our compiler to rename all the main step in a task as "main". Argo did the container renaming in the execution level since Argo only allows one container as "main" and the rest are initContainer. Whereas Tekton only has the concept of steps but not initContainer. 

@ckadner what do you think?

**Additional information:**

Tomcli:
Resolves #154 

Since Tekton allows multiple step spec in a task whereas Argo only allows one container spec per task, it's not obvious for Tekton to tell which step is from the main ContainerOp. Therefore, we want to rename all the containerOp step name to "main" to distinguish between the initContainers and main step.

This has no impact to our existing examples because we just change the container name for logging with the API.

**Environment tested:**

* Tekton Version (use `tkn version`): 0.11.3

### DIFF
--- a/sdk/FEATURES.md
+++ b/sdk/FEATURES.md
@@ -9,6 +9,7 @@ Below are the list of features that are currently available in the KFP Tekton co
     + [Timeout for Tasks and Pipelines](#timeout-for-tasks-and-pipelines)
     + [RunAfter](#runafter)
     + [Input Parameters](#input-parameters)
+    + [ContainerOp](#containerop)
 - [Pipeline DSL features with custom Tekton implementation](#pipeline-dsl-features-with-custom-tekton-implementation)
   * [Features with same behavior as Argo](#features-with-same-behavior-as-argo)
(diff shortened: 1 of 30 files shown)


drewbutlerbb4:
Can we do the same for Task templates created for ResourceOps

In the compiler: [_op_to_template.py](https://github.com/Tomcli/kfp-tekton/blob/main-task/sdk/python/kfp_tekton/compiler/_op_to_template.py#L130)

In the testdata: [resourceop_basic.yaml](https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/testdata/resourceop_basic.yaml#L67)

Tomcli:
Thanks @drewbutlerbb4, I addressed the comment.

Issue #150, opened 2020-05-18
Compiler issues reported by Watson


animeshsingh:
/kind bug

Reported by Michalina  @udiknedormin

* identifiers are not checked for length, some of them fail on length check (63 characters is the limit)
* not all identifiers are escaping `_` signs (`_` -> `-`)
* no big data fix by default, instead it's hidden behind a flag, triggering very cryptic Tekton error
  * error message is "no such an output", which contradicts the definition
  * origins in Tekton-Pipelines way of handling outputs from termination-message
  * termination message has length cap, which can be easily exceeded for big inputs/outputs
  * Tekton-Pipelines doesn't do any error handling when parsing json from termination message, just parses empty list on error
  * messages that are too long don't seem to follow one pattern: they can have their end cut, their beginning cut or both
* big data fix still uses `results` to pass the whole value, not just the path, so the parsing error is still happening
* big data fix makes results multiplied: output A of step X is present in both X and `copy-artifacts` step injected by the compiler, Tekton takes both
* `copy-artifacts` only copies them out, for output artefacts, but doesn't copy them to the by-artifact inputs of consumers

ckadner:
I think most of the reported problems should be related to these open issues:

- [ ] Validate identifiers #151
- [ ] Fix big data passing #63
- [ ] Verify generated Tekton YAML #137

animeshsingh:
cc @fenglixa 

@ckadner can you connect with @fenglixa and split the effort?

fenglixa:
I am tring to work on "Re-implement the fix_big_data_passing function" #63 firstly.

Also other issues inside need to be addressed: 

- [ ] Exit handler support #85
- [ ] messages that are too long don't seem to follow one pattern: they can have their end cut, their beginning cut or both
- [ ] Uses results to pass the whole value, not just the path, so the parsing error is still happening
- [ ] big data fix makes results multiplied: output A of step X is present in both X and copy-artifacts step injected by the compiler, Tekton takes both
- [ ] copy-artifacts only copies them out, for output artefacts, but doesn't copy them to the by-artifact inputs of consumers

ckadner:
> I am trying to work on "Re-implement the fix_big_data_passing function" #63 firstly.

@fenglixa -- I assigned #63 to you. Thanks for your help!

fenglixa:
Connected with @Udiknedormin on slack. This issue was addressed.  Seems it could be closed now.

/close

PR #158, opened 2020-05-21
Sanitize Kubernetes Names

### LINKED ISSUE #151: Validate identifiers
- identifiers have a 63 character limit
- consistently replace under score characters `_` with dashes `-` (`_` -> `-`)

ckadner:
Resolves #151 

`sanitize_k8s_name` cleans and converts the names in the workflow according to [K8s name and label spec](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names)

**Environment tested:**
* Python Version (use `python --version`): `3.7`
* Tekton Version (use `tkn version`): `0.11`
* Kubernetes Version (use `kubectl version`): `0.15`
* OS (e.g. from `/etc/os-release`): `MacOS 10.15.3`

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_op_to_template.py
+++ b/sdk/python/kfp_tekton/compiler/_op_to_template.py
@@ -21,7 +21,7 @@ from collections import OrderedDict
 from typing import List, Text, Dict, Any
 
 from kfp import dsl
-from kfp_tekton.compiler._k8s_helper import convert_k8s_obj_to_json
+from kfp_tekton.compiler._k8s_helper import convert_k8s_obj_to_json, sanitize_k8s_name
 from kfp.compiler._op_to_template import _process_obj, _inputs_to_json, _outputs_to_json
 from kfp.dsl import ArtifactLocation
 from kfp.dsl._container_op import BaseOp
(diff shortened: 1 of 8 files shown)


ckadner:
@Tomcli -- when you get a chance, could you help review this PR? Thanks :-)

@Udiknedormin -- please let me know if this PR addresses the identifier issues when compiling and running your pipelines (#150, #151). Thank you.

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -490,9 +490,15 @@ def _op_to_template(op: BaseOp, enable_artifacts=False):
     if processed_op.pod_annotations or processed_op.pod_labels:
         template.setdefault('metadata', {})  # Tekton change, don't wipe out existing metadata
         if processed_op.pod_annotations:
-            template['metadata']['annotations'] = processed_op.pod_annotations
+            template['metadata']['annotations'] = {
+                sanitize_k8s_name(key, True, True, True, 253): value
```
Tomcli, line 494: nit: There are some True flags for allowing dot/underscore/slash. Maybe is better to use [python keyword arguments](https://www.w3schools.com/python/gloss_python_function_keyword_arguments.asp) to explicitly show the special characters we are allowing here?

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -490,9 +490,15 @@ def _op_to_template(op: BaseOp, enable_artifacts=False):
...
+                for key, value in processed_op.pod_annotations.items()
+            }
         if processed_op.pod_labels:
-            template['metadata']['labels'] = processed_op.pod_labels
+            template['metadata']['labels'] = {
+                sanitize_k8s_name(key, True, True, True, 253): sanitize_k8s_name(value, True, True)
```
Tomcli, line 499: nit: same as above.

sdk/python/kfp_tekton/compiler/_op_to_template.py
ckadner: I wanted to keep those code blocks as short as possible, but you are right, it is easier to read and prevents errors in the future if we use explicit parameter names

sdk/python/kfp_tekton/compiler/_op_to_template.py
ckadner: I pushed another commit. Now using keyword arguments where there were more than one `True`/`False` flags in a row

ckadner:
Thanks @Tomcli

Issue #165, opened 2020-06-01
Update conditional feature with the catalog condition approach


Tomcli:
/kind feature

**Description:**
We want to rewrite the Tekton condition CRD to be more genetic for KFP. This way we can treat our condition as a "catalog" and let pipelineRun refer to it without passing multiple yamls.

TODOs:
1. Generalize the current condition into a more catalog task design.
2. Have a flag for compiler to opt-out the condition yaml since it will be available in KFP by default.
3. Update our Tekton set up to also include the "catalog condition" we created in Task 1.

**Additional information:**
[Miscellaneous information that will assist in solving the issue.]

PR #167, opened 2020-06-02
Update conditional feature with the catalog condition approach

drewbutlerbb4:
Addresses task 1 of #165: Generalize the current condition into a more catalog task design.

Creates a "super-condition" which all other condition extend by passing in the parameters necessary for individual conditions

### DIFF
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -42,7 +42,7 @@ def _get_super_condition_template():
 
   python_string = 'python -c \'import sys\ninput1=str.rstrip(sys.argv[1])\ninput2=str.rstrip(sys.argv[2])\n' \
     + 'try:\n  input1=int(input1)\n  input2=int(input2)\nexcept:\n  input1=str(input1)\n' \
-    + 'sys.exit(0) if (input1 $(params.condition) input2) else sys.exit(1)\' \'$(params.operand1)\' \'$(params.operand2)\''
+    + 'sys.exit(0) if (input1 $(params.operator) input2) else sys.exit(1)\' \'$(params.operand1)\' \'$(params.operand2)\''
 
   # TODO Change to tekton_api_version once Conditions are out of v1alpha1
   template = {
@@ -55,7 +55,7 @@ def _get_super_condition_template():
       'params': [
         {'name': 'operand1'},
         {'name': 'operand2'},
-        {'name': 'condition'}
+        {'name': 'operator'}
       ],
       'check': {
         'args': [python_string],
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -40,9 +40,16 @@ from kfp_tekton import tekton_api_version
 
 def _get_super_condition_template():
 
-  python_string = "python -c 'import sys;input1=str.rstrip(sys.argv[1]);input2=str.rstrip(sys.argv[2]);" \
-    + "try:  input1=int(input1);  input2=int(input2);except:  input1=str(input1);" \
-    + "sys.exit(0) if (input1 $(params.operator) input2) else sys.exit(1)' '$(params.operand1)' '$(params.operand2)'"
+  python_string = textwrap.dedent('''\
+    'import sys
+    input1=str.rstrip(sys.argv[1])
+    input2=str.rstrip(sys.argv[2])
+    try:
+      input1=int(input1)
+      input2=int(input2)
+    except:
+      input1=str(input1)
+    sys.exit(0) if (input1 $(params.operator) input2) else sys.exit(1)' ''')
 
   # TODO Change to tekton_api_version once Conditions are out of v1alpha1
   template = {
@@ -58,8 +65,7 @@ def _get_super_condition_template():
         {'name': 'operator'}
       ],
       'check': {
-        'args': [python_string.replace(';','\n')],
-        'command': ['sh', '-c'],
+        'script': 'python -c ' + python_string + "'$(params.operand1)' '$(params.operand2)'",
         'image': 'python:alpine3.6',
       }
     }
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -196,7 +196,7 @@ class TektonCompiler(Compiler):
       template['spec']['params'] = [
         {'name': 'operand1', 'value': operand1_value},
         {'name': 'operand2', 'value': operand2_value},
-        {'name': 'condition', 'value': str(condition.operator)}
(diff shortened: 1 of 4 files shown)


Tomcli:
@drewbutlerbb4 can you create a condition template similar to this and put it under `tekton-catalog`? https://github.com/kubeflow/kfp-tekton/blob/master/tekton-catalog/kubectl-wrapper/deploy/taskrun.yaml 

It would be nice if you can also add a short readme to describe each parameter in the catalog condition

drewbutlerbb4:
Thanks @Tomcli, I have made the requested changes

Tomcli:
Thanks @drewbutlerbb4 
@ckadner do you want to hold this PR until you merge your pipelineRun changes?

tekton-catalog/condition/README.md
```diff
@@ -0,0 +1,8 @@
+# condition
+Conditional statements are frequently used in Tekton and as such a catalog condition has been provided as a simple way to
```
animeshsingh, line 2:
```suggestion
Conditional statements are frequently used in Kubeflow Pipelines, and as such a Tekton Catalog Task for condition has been provided as a simple way to 
```

tekton-catalog/condition/README.md
```diff
@@ -0,0 +1,8 @@
+# condition
+Conditional statements are frequently used in Tekton and as such a catalog condition has been provided as a simple way to 
+invoke conditions.
```
animeshsingh, line 3:
```suggestion
invoke conditions. This is pending the support for formal Conditions spec in Tekton V1beta1 API, which is being developed according the [design specs in Tekton community](https://docs.google.com/document/d/1kESrgmFHnirKNS4oDq3mucuB_OycBm6dSCSwRUHccZg/edit).
```

ckadner:
> Thanks @drewbutlerbb4
> @ckadner do you want to hold this PR until you merge your `pipelineRun` changes?

No. I don't think this PR should be held up.

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -363,22 +373,28 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
...
     # generate task and condition reference list for the Tekton Pipeline
     condition_refs = {}
     task_refs = []
-    for template in templates:
+    templates = []
+    condition_added = False
+    for template in raw_templates:
       if template['kind'] == 'Condition':
+        if not condition_added:
+          templates.append(_get_super_condition_template())
```
Tomcli, line 386: nit: Add a TODO comment here for opt-out the condition_template in the future.

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -38,6 +38,36 @@
 from kfp_tekton import tekton_api_version
 
 
+def _get_super_condition_template():
+
+  python_string = "python -c 'import sys;input1=str.rstrip(sys.argv[1]);input2=str.rstrip(sys.argv[2]);" \
```
Tomcli, line 43: Another way you can make it cleaner is to use the condition script instead of args and commands.
https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/conditional-pipelinerun-with-same-condition-refer.yaml#L13-L21

Then you can use the textwrap library to put things into a single string.

Tomcli: https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/kfp_tekton/compiler/_op_to_template.py#L308-L311

drewbutlerbb4:
Thanks @animeshsingh @Tomcli @ckadner, I have amended based on your suggestions.

@Tomcli @ckadner are the changes made agreeable to you?

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -40,9 +40,16 @@
 
 def _get_super_condition_template():
 
-  python_string = 'python -c \'import sys\ninput1=str.rstrip(sys.argv[1])\ninput2=str.rstrip(sys.argv[2])\n' \
-    + 'try:\n  input1=int(input1)\n  input2=int(input2)\nexcept:\n  input1=str(input1)\n' \
-    + 'sys.exit(0) if (input1 $(params.operator) input2) else sys.exit(1)\' \'$(params.operand1)\' \'$(params.operand2)\''
+  python_string = textwrap.dedent('''\
```
ckadner, line 43: Thanks @drewbutlerbb4  this looks much nicer :-)  ...but I would call it `python_script` as opposed to `python_string`

Tomcli: it looks like this new script created an error, here's the log
```
+ python -c import sys
input1=str.rstrip(sys.argv[1])
input2=str.rstrip(sys.argv[2])
try:
  input1=int(input1)
  input2=int(input2)
except:
  input1=str(input1)
sys.exit(0) if (input1 == input2) else sys.exit(1) tails
 heads
```

Tomcli: maybe is better to rewrite the static template to have it working first. Then we can port the string from the template into this code.

drewbutlerbb4: I'm not seeing that error on my end, would you mind giving more information @Tomcli? This issue is at compile time or runtime?

Tomcli: it happened during runtime, I'm using the same unit test example in this PR. My Tekton version is 0.12.0

Tomcli: Oh sorry I think Tekton intended to show errors. The pipeline status said "completed" instead of "Succeeded", so it was breaking my API code.

tekton-catalog/condition/condition.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Condition
+metadata:
+  name: super-condition
+spec:
+  check:
+    args:
```
Tomcli, line 7: The script looks good to me now. Can you also update the static template to follow the same design?

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: We probably need to update the API and UI code to catch these new status.

drewbutlerbb4: Sure, I can update this for the UI. 'completed' is not a valid status for kfp there don't seem to be any direct equivalents but 'Terminated' seems to be the closest.

Tomcli:
thanks @drewbutlerbb4 
/lgtm

ckadner:
@drewbutlerbb4 @Tomcli -- did we ever document how to install the `super-condition`?   (@moficodes)

PR #169, opened 2020-06-04
Fix for big data passing

### LINKED ISSUE #63: Fix big data passing

fenglixa:
Resolves #63
And related to #150

Big data should be read from files and written to files.

The paths for the input and output files are chosen by the system and are passed into the function (as strings).

Use the InputPath parameter annotation to tell the system that the function wants to consume the corresponding input data as a file. The system will download the data, write it to a local file and then pass the path of that file to the function.

Use the OutputPath parameter annotation to tell the system that the function wants to produce the corresponding output data as a file. The system will prepare and pass the path of a file where the function should write the output data. After the function exits, the system will upload the data to the storage system so that it can be passed to downstream components.

Notes: As we used 'workspaces' in tekton pipelines to handle the bigger data processing, the complier will generator the PVC definations,it needs the volume to store the data. User need to create volume manually, or enable dynamic volume provisioning, refer to the link of: https://kubernetes.io/docs/concepts/storage/dynamic-provisioning

- Implementation:

- [x] Index the PIPELINEs to understand how data is being passed and which inputs/outputs
       are connected to each other.
- [x] Search for direct data consumers in container/resource templates and some
       PIPELINE task attributes (e.g. conditions and loops) to find out which inputs
       are directly consumed as parameters/artifacts.
- [x] Propagate the consumption information upstream to all inputs/outputs all
       the way up to the data producers.
- [x] Convert the inputs, outputs based on how they're consumed downstream.
- [x] Use workspaces instead of result and params for bigger data passing.
- [x] Added workspaces to tasks, pipelines, pipelineruns, if the parmas is bigger data.
- [x] A PVC named pipelinerun name will be created if bigger data passing, as workspaces need to use it.
- [x] Document about 'How to use it'

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -13,17 +13,18 @@
 # limitations under the License.
 
 from typing import List, Text, Dict, Any
+# from kfp.compiler._data_passing_rewriter import fix_big_data_passing
+import copy
+import json
+import re
+from typing import Optional, Set
+from . import _op_to_template
 
 
-def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any]]:  # Tekton change signature
+def fix_big_data_passing(
+    workflow: List[Dict[Text, Any]]
+) -> List[Dict[Text, Any]]:  # Tekton change signature
     """
-    Currently this function does not do anything.
-
-    TODO: evaluate if an implementation similar to the upstream function in KFP
-          is required
-
-    Documentation from KFP:
-
     fix_big_data_passing converts a workflow where some artifact data is passed
     as parameters and converts it to a workflow where this data is passed as
     artifacts.
@@ -35,12 +36,12 @@ def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any
 
     Motivation:
 
-    DSL compiler only supports passing Argo parameters.
+    DSL compiler only supports passing Tekton parameters.
     Due to the convoluted nature of the DSL compiler, the artifact consumption
     and passing has been implemented on top of that using parameter passing.
     The artifact data is passed as parameters and the consumer template creates
     an artifact/file out of that data.
-    Due to the limitations of Kubernetes and Argo this scheme cannot pass data
+    Due to the limitations of Kubernetes and Tekton this scheme cannot pass data
     larger than few kilobytes preventing any serious use of artifacts.
 
     This function rewrites the compiled workflow so that the data consumed as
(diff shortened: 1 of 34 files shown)


fenglixa:
/cc @animeshsingh @ckadner @Tomcli 
Thanks very much!

Udiknedormin:
@fenglixa After merging and applying `generate_pipelinerun=True` flag (as per assertion's error message), some tasks get their parameters removed despite using them. Because of that, the task refers to a non-existing parameter and pipeline validation fails on "passing unnecessary input parameter".

I will provide you with details via internal communication channels.

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -451,16 +451,19 @@ def _op_to_template(op: BaseOp, enable_artifacts=False):
...
+        # for artifact in inputs['artifacts']:
+        #     if 'raw' in artifact:
+        #         copy_inputs_step['script'] += 'echo -n "%s" > %s\n' % (artifact['raw']['data'], artifact['path'])
+        #     mount_path = artifact['path'].rsplit("/", 1)[0]
+        #     if mount_path not in mounted_param_paths:
+        #         _add_mount_path(artifact['name'], artifact['path'], mount_path,
+        #                         volume_mount_step_template, volume_template, mounted_param_paths)
+        # template['spec']['steps'] = _prepend_steps([copy_inputs_step], template['spec']['steps'])
+        # _update_volumes(template, volume_mount_step_template, volume_template)
+        # Francis Todo: above info may need to be moved to bigger data passing logic
```
Tomcli, line 464: Do we still need to keep these commented code since you already moved them to the bigger data passing logic?

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -69,7 +69,7 @@ def _get_resourceOp_template(op: BaseOp,
     template = {
         'apiVersion': tekton_api_version,
         'kind': 'Task',
-        'metadata': {'name': name},
+        'metadata': {'name': name, 'labels': {'app': 'resourceOp'}},
```
Tomcli, line 72: @ckadner Does the new pipelineRun design allow adding labels to individual task?

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -52,14 +53,615 @@ def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any
...
+# TODO: Remove PVC if Tekton version > = 0.12, use 'volumeClaimTemplate' instead
+def create_pvc(pr: str) -> dict:
+    pvc = {
+        "apiVersion": "v1",
+        "kind": "PersistentVolumeClaim",
+        "metadata": {
+            "name": pr
+        },
+        "spec": {
+            "accessModes": ["ReadWriteOnce"],
```
Tomcli, line 617: With RWO, tasks in different k8s node won't able to write/mount at the same time. Is it the same behavior in Argo?

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -52,14 +53,615 @@ def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any
...
+        del task['spec']['artifacts']
+
+    return task
+
+
+# Create pvc for pipelinerun if bigger data passing.
+# As we used workspaces in tekton pipelines which depends on it.
+# User need to create PV manually, or enable dynamic volume provisioning, refer to the link of:
+# https://kubernetes.io/docs/concepts/storage/dynamic-provisioning
+# TODO: Remove PVC if Tekton version > = 0.12, use 'volumeClaimTemplate' instead
```
Tomcli, line 608: The current design will use the same pvc for all the pipelinerun. Did Argo store them as "artifacts" in the original design?

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: Oh, Thank you. Tomcli. I forgot delete it.

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: I added a labels for all the resourceOp tasks to distinguish resource and normal container. @ckadner, please let me know if not meet your requirements, or provide me a suggestion. Thanks! : )

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
fenglixa: Since I added the `runAfter` in the workspaces tasks. So there should be no situation of writing same file at same time

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
fenglixa: I difined different files for the bigger_data params,  I think we should reduce the pvc/pv number creating for one pipeline run, which mean reduce the effort for user of manual steps, there is no reason to create several pvcs for all the params.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
fenglixa: OK， get it,  seems we should use RWX, but not RWO, will double check the behavior in Argo.

fenglixa:
Thanks @Udiknedormin 's cooperation. Fixed the issue Michalina telled me, it's my code error. which was fixed by me just now.
Another issue we enconterred is an old issue which related to volumeMounts. I will provide another PR later to fix it.

animeshsingh:
Thanks @fenglixa - great job!!!! 

@Tomcli @Udiknedormin @ckadner any final comments, or is good to go?

OWNERS
```diff
@@ -2,6 +2,7 @@ approvers:
  - animeshsingh
  - ckadner
  - Tomcli
+ - fenglixa
```
ckadner, line 5: @fenglixa @animeshsingh -- not debating merit, but I think that a promotion to become "committer" should be part of a separate PR

sdk/python/kfp_tekton/compiler/_op_to_template.py
ckadner: no, with embedded `taskSpec` `metedata` is not allowed as far as I know

@afrittoli -- do you have any advice about enabling `metadata` for embedded `taskSpec`s?

OWNERS
animeshsingh: Definitely, you deserve to be an approver here @fenglixa. Let's just do it in a separate PR using the community norms, so that we can approve and celebrate!!!

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -52,14 +53,615 @@ def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any
...
+
+    # Need to confirm:
+    # I didn't find the use cases to support workflow parameter consumed as artifacts downstream in tekton.
+    # Whether this case need to be supporting?
+
+    clean_up_empty_workflow_structures(workflow)
     return workflow
+
+
+# Replaced '_' to '-' to make the format in tekton unified for params, artifacts, outputs.
```
ckadner, line 416: we have a general method that ["sanitizes" Kubernetes names](https://github.com/kubeflow/kfp-tekton/blob/2c68de4/sdk/python/kfp_tekton/compiler/_k8s_helper.py#L20) which should be used instead of `replace('_', '-')`

...which should have been invoked already by [`_sanitize_and_inject_artifact()`](https://github.com/kubeflow/kfp-tekton/blob/d806bc8/sdk/python/kfp_tekton/compiler/compiler.py#L534). If some names were missed it should be added there.

OWNERS
fenglixa: Thanks @ckadner, @animeshsingh , I will log separate PR for it.

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: Drop the lables for resourceOp. Resolve this issue

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -355,7 +355,7 @@ def _process_base_ops(op: BaseOp):
 
     # map param's (unsanitized pattern or serialized str pattern) -> input param var str
     map_to_tmpl_var = {
-        (param.pattern or str(param)): '$(inputs.params.%s)' % param.full_name  # Tekton change
+        (param.pattern or str(param)): '$(inputs.params.%s)' % param.full_name.replace('_', '-')  # Tekton change
```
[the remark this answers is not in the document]
fenglixa: I cannot use sanitize_k8s_name, since it will lower the letter if not allow_capital_underscore, which will not along to the outputs and other params.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -52,14 +53,615 @@ def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any
 
     Implementation:
 
-    1. Index the DAGs to understand how data is being passed and which inputs/outputs
+    1. Index the PIPELINEs to understand how data is being passed and which inputs/outputs
```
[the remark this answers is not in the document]
fenglixa: Thanks @ckadner , Addressed

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -52,14 +53,615 @@ def fix_big_data_passing(workflow: List[Dict[Text, Any]]) -> List[Dict[Text, Any
...
+       PIPELINE task attributes (e.g. conditions and loops) to find out which inputs
        are directly consumed as parameters/artifacts.
     3. Propagate the consumption information upstream to all inputs/outputs all
        the way up to the data producers.
-    4. Convert the inputs, outputs and arguments based on how they're consumed
-       downstream.
+    4. Convert the inputs, outputs based on how they're consumed downstream.
+    5. Use workspaces instead of result and params for bigger data passing.
+    6. Added workspaces to tasks, pipelines, pipelineruns, if the parmas is bigger data.
+    7. Convert "_" to "-" in the params and artifacts.
```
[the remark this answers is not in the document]
fenglixa: Thanks @ckadner , Yes, I tested, we have to covert _ to - for the params, since the results value didn't support it, and we already coverted it, this will lead to inputs and outputs inconsistency. This was tested out for my big data passing cases, and also reproduced for watson cases.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
fenglixa: Hi @ckadner ,
I tried to add the name and use _sanitize_and_inject_artifact() almost whole day yesterday, but failed, the major reasons:
1.  The params and artifacts cannot be converted before workflow and templates created. otherwise, some value will be mis-matched, it will impact to other features, e.g: loop, conditions. etc...
2. _sanitize_and_inject_artifact() is not suitable to use after workflow created, because a. `_sanitize_and_inject_artifact` will remove the symbol of "$,{,}, etc..", b.  `allow_capital_underscore` will convert  capital and underscore in the same time

Personally, I think I can provide another PR separately, try to unified the params in bigger data and _sanitize_and_inject_artifact. How do you think?

sdk/python/tests/compiler/testdata/artifact_location.yaml
```diff
@@ -20,7 +20,7 @@ spec:
   params:
   - name: bucket
   - name: namespace
-  - name: secret_name
+  - name: secret-name
```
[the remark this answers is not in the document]
fenglixa: Thanks @ckadner, yes capital letters and underscore characters are support for parameter name, but not for results value, and we already coverted it in results value, this will lead to inputs and outputs inconsistency. It was tested out for my big data passing cases, and also reproduced for watson cases.

fenglixa:
> Any reason to use the term "big**ger**_data_passing" vs "big_data_passing" as used in KFP?

Good quetions, to avoid any confuse, I updated it all as `big_data_passing` which same as KFP. Thanks @ckadner

fenglixa:
Addressed the comments and fixed the issues for watson case, big data passing is working for old feature, big data, and also Watson pipelines now.
@Tomcli @ckadner @Udiknedormin , Is it OK for you now? thanks!

Udiknedormin:
@fenglixa Works fine on our side. Thank you!

Issue #163, opened 2020-05-23
Remove ArtifactLocation support when migrating to KFP 0.5.1


Tomcli:
/kind bug

**What steps did you take and what happened:**
KFP 0.5.1 has officially removed the ArtifactLocation support. So we need to remove the ArtifactLocation in kfp-tekton when we migrate to KFP 0.5.1. Otherwise it will produce an `ImportError`.

**What did you expect to happen:**

**Additional information:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

* Python Version (use `python --version`):
* SDK Version: 
* Tekton Version (use `tkn version`):
* Kubernetes Version (use `kubectl version`):
* OS (e.g. from `/etc/os-release`):

fenglixa:
@Tomcli I am handling input/output artifacts in `Fix big data passing` task # 63, which will hanlde all kinds of input/output artifacts.
So, could you please hold on this task, as it's part of my on-going task now. Thanks!

Tomcli:
@fenglixa no problem, I opened this issue as a TODO for the KFP version upgrade.

fenglixa:
Thanks @Tomcli, ArtifactLocation could be removed now, since issue #63 was done.  Not sure whether you will handle it later or any helps from my side do you need?

Tomcli:
@fenglixa, I can take care of this. If you have time, can you take a look how we can warp our container logs and send them to the object storage? https://github.com/kubeflow/kfp-tekton/issues/161

fenglixa:
> @fenglixa, I can take care of this. If you have time, can you take a look how we can warp our container logs and send them to the object storage? #161

OK, sure.

PR #170, opened 2020-06-06
Add artifact annotation for metadata tracking

### LINKED ISSUE #168: Add artifact input and output annotations for artifact tracking
/kind feature

**Description:**
Argo controller can embedded the input and output artifacts into the pod annotation to achieve artifact tracking. However, Tekton doesn't have the concept of artifacts (since pipelineResource is still in alpha), we need to add the artifacts information within the sdk compiler in order to achieve similar feature. 

**Additional information:**

Tomcli:
Resolves #168 

Add annotation to support artifacts from input/output parameters. Input artifacts with big data passing is ignored because we are pending #169. 

Change the output artifact location and add annotations for metadata tracking. The design of this approach is documented in this google docs
https://docs.google.com/document/d/1nx63yZzANk8yDAm_Cwx_UHu-SVFrg73GVeshhtaYNsM/edit#heading=h.d1ddt7vobtmu

Major changes:
- Moved `PODNAME` to `PIPELINETASK` as the secondary directory for S3 artifact key.
- Follow the [Argo standard](https://github.com/argoproj/argo/blob/6e5dd2e19a3094f88e6f927f786f866eccc5f500/workflow/controller/workflowpod.go#L877-L878) to compress artifacts into tgz file
- Add default pipelineRun annotation for metadata tracking. Default to `{}` to indicate the pipeline has no artifact.

**Environment tested:**

* Tekton Version (use `tkn version`): 1.12.1

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_op_to_template.py
+++ b/sdk/python/kfp_tekton/compiler/_op_to_template.py
@@ -368,7 +368,7 @@ def _process_base_ops(op: BaseOp):
     return op
 
 
-def _op_to_template(op: BaseOp, tekton_compiler):
+def _op_to_template(op: BaseOp, pipelinerun_output_artifacts={}, enable_artifacts=False):
     """Generate template given an operator inherited from BaseOp."""
 
     # initial local variables for tracking volumes and artifacts
@@ -400,7 +400,7 @@ def _op_to_template(op: BaseOp, tekton_compiler):
                     path=path,
                     key='runs/$PIPELINERUN/$PIPELINETASK/' + name))
             for name, path in output_artifact_paths.items()
-        ] if tekton_compiler.enable_artifacts else []
+        ] if enable_artifacts else []
 
         # workflow template
         container = convert_k8s_obj_to_json(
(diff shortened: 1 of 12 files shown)


sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -366,7 +368,7 @@ def _process_base_ops(op: BaseOp):
     return op
 
 
-def _op_to_template(op: BaseOp, enable_artifacts=False):
+def _op_to_template(op: BaseOp, tekton_compiler):
```
ckadner, line 371: All the op-to-template conversions here are static in a sense that they do not rely on the compiler instance. Any compiler instance specific template manipulations are in the `TektonCompiler` class. Can the output artifact tracking annotation code move to the `TektonCompiler`, or, alternatively we pass the `output_artifacts` into this `_op_to_template` method?

Tomcli: I changed the arguments to only pass the necessary objects. I also added a new test case because we are going to remove the `artifact_location` test after Feng's PR. (cc #163)

PR #177, opened 2020-06-10
Consolidate the value of results to avoid underscore convering in params

fenglixa:
Capital letters and underscore characters are supported for parameter name, but not for results in tekton pipelines, and we already coverted the results name/values before workflow created, this will lead to inputs and outputs inconsistency, which lead cases failed. This PR was consolidate the results value converting, by deleting the converting before workflow created, to avoid any underscore convering in params.

/cc @ckadner

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -19,6 +19,7 @@ import json
 import re
 from typing import Optional, Set
 from . import _op_to_template
+from kfp_tekton.compiler._k8s_helper import sanitize_k8s_name
 
 
 def fix_big_data_passing(
@@ -96,11 +97,6 @@ def fix_big_data_passing(
     pipelinerun_templates = [
         template for template in workflow if template['kind'] == 'PipelineRun'
     ]
-
-    # we need to make the format in tekton unified for params, artifacts, outputs.
-    container_templates = unify_container_params(container_templates)
-    pipeline_templates = unify_pipeline_params(pipeline_templates)
-    pipelinerun_templates = unify_pipelinerun_params(pipelinerun_templates)
 
     # 1. Index the pipelines to understand how data is being passed and which
     #  inputs/outputs are connected to each other.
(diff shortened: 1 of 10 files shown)


sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -412,6 +422,20 @@ def mark_upstream_ios_of_output(template_output, marked_inputs,
...
 
+            # tekton results doesn't support underscore
+            for argument in task['params']:
+                argument_value = argument.get('value')
+                argument_placeholder_parts = deconstruct_tekton_single_placeholder(
+                    argument_value)
+                if argument_placeholder_parts and argument_placeholder_parts[
+                        0] == 'tasks' and argument_placeholder_parts[
+                            1] and argument_placeholder_parts[
+                                2] and argument_placeholder_parts[3]:
```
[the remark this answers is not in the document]
fenglixa: Updated

fenglixa:
> so this PR will only sanitize underscores for Tekton results only.
Yes, this PR will sanitize Tekton results only, to avoid any params converting.

fenglixa:
Thanks @ckadner , I updated the code based on your comments, could you please check whether it's OK for you now? Thanks!

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -412,6 +422,21 @@ def mark_upstream_ios_of_output(template_output, marked_inputs,
...
+            # tekton results doesn't support underscore
+            for argument in task['params']:
+                argument_value = argument.get('value')
+                argument_placeholder_parts = deconstruct_tekton_single_placeholder(
+                    argument_value)
+                if argument_placeholder_parts \
+                    and argument_placeholder_parts[0] == 'tasks' \
+                    and argument_placeholder_parts[1] \
+                    and argument_placeholder_parts[2] \
+                        and argument_placeholder_parts[3]:
```
ckadner, line 434: I meant, replace that:

```Python
                if argument_placeholder_parts \
                    and argument_placeholder_parts[0] == 'tasks' \
                    and argument_placeholder_parts[1] \
                    and argument_placeholder_parts[2] \
                        and argument_placeholder_parts[3]:
```

with this:
```Python
                if len(argument_placeholder_parts) == 4 \
                        and argument_placeholder_parts[0] == 'tasks':
```

And change the method `deconstruct_` from:

```Python
def deconstruct_tekton_single_placeholder(s: str) -> List[str]:
    if not re.fullmatch('\\$\\([-._a-zA-Z0-9]+\\)', s):
        return None
    return s.lstrip('$(').rstrip(')').split('.')
```

to:
```Python
def deconstruct_tekton_single_placeholder(s: str) -> List[str]:
    if not re.fullmatch(r'\$\([-._a-zA-Z0-9]+\)', s):  # use an R-string
        return []                                      # return empty list instead of None
    return s.lstrip('$(').rstrip(')').split('.')
```

fenglixa: OK, got your mean, thanks.

fenglixa:
Thanks @ckadner, please help check again.

PR #187, opened 2020-06-13
Update E2E tests

ckadner:
Updating a few test logs after the latest artifact changes.

- Break up test cases into 2 sub-tests:
  1. Validate Tekton YAML
  2. Verify PipelineRun logs
- Delete tasks before running pipeline as some pipelines re-use same task names but have a slightly different implementation causing unintended results
- Normalize UUID before comparing logs

**Environment tested:**

* Python Version (use `python --version`): `3.7.6`
* Tekton Version (use `tkn version`): `v20200611-1caf5bffe1` (nightly build)
* Kubernetes Version (use `kubectl version`): `0.16.10`
* OS (e.g. from `/etc/os-release`): MacOS Catalina

/cc @Tomcli


Tomcli:
/lgtm
Thanks @ckadner, is the e2e test checking the logs in the exact order? For pipelines that have parallel tasks will have in different log order each time.

ckadner:
> /lgtm
> Thanks @ckadner, is the e2e test checking the logs in the exact order? For pipelines that have parallel tasks will have in different log order each time.

[`    def _sanitize_log(self, log) -> str:`](https://github.com/ckadner/kfp-tekton/blob/46649da19ca00c505c3e44c43457e8a75b143dbb/sdk/python/tests/compiler/compiler_tests_e2e.py#L202-L240):
```Python
        # sort lines since parallel tasks produce log lined in unpredictable order
        sanitized_log = "\n".join(sorted(sanitized_log.splitlines()))
```

Tomcli:
I see, thanks @ckadner 
/approve

PR #210, opened 2020-07-02
Make unit tests run with Python 3.5

ckadner:
**Problem:**

Python before `3.6` did not preserve insert order for dictionaries, which causes the generated YAML to be equivalent, yet it can not reliably be compared to the "golden" YAML via `unittest.assertDictEqual`.

**Solution:**

* Sort `dict`s (and `list`s) in unit tests if the Python test runtime predates version 3.6.0
* Ensure string-serialized values are sorted via `json.dumps(..., sort_keys=True)` in the compiler code
* Add unit tests for Python 3.5 to Travis/CI
* Exclude Katib workflow from py35 unit tests as it uses (string-serialized) dictionaries containing `dsl.PipelineParam` objects which can't be JSON-serialized so using `json.dumps(sort..)` is not an alternative

**Environment tested:**

* Python Version (use `python --version`): `3.5`, `3.6`, `3.7` (via `tox`)

/cc @Tomcli

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,14 @@
 
 matrix:
   include:
+    - name: "Unit tests, Python 3.5"
+      language: python
+      python: "3.5"
+      env: TOXENV=py35
+      install:
+        - python3 -m pip install -e sdk/python
+      script:
+        - make unit_test
     - name: "Unit tests, Python 3.6"
       language: python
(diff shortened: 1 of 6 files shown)


ckadner:
Hi @Tomcli -- I rebased to resolve the merge conflicts

PR #215, opened 2020-07-03
Verify Markdown files have current Table of Contents

ckadner:

* Update `tools/mdtoc.sh` script to remove colons from anchor links
* Add `check_mdtoc` target to `Makefile`
* Add `make check_mdtoc` to Travis/CI

**Environment tested:**

* Python Version: `3.7.5`
* Tekton Version: --
* Kubernetes Version: --
* Operating System: MacOS `10.15.5`

/cc @Tomcli


.travis.yml
```diff
@@ -37,15 +37,19 @@ matrix:
...
       script:
         - make lint
+    - name: "Verify source files contain the license header"
+      language: bash
+      script:
+        - make check_license
+    - name: "Verify Markdown files have current table of contents"
+      language: bash
+      script:
+        - make check_mdtoc
```
Tomcli, line 52: Can we group some these tests (e.g. license and mdtoc check) into one job before we start overwhelming the Travis CI quota?

ckadner: What is the quota we have with Travis? Until we hit a limit I would like to keep them separate. Maybe there is a way to share common setup between Travis jobs but still report separate logical jobs?

Tomcli: Right now it should be fine. For open source I think we are allowed to run up to 200 jobs in parallel per repo. So if we have more jobs, less commits/PR can be tested in parallel.

Tomcli: This is something to put in mind as we creating more tests. For now, it LGTM

PR #216, opened 2020-07-03
Upgrade to Tekton version 0.14.0

ckadner:
Tekton has just released version [`0.14.0`](https://github.com/tektoncd/pipeline/releases/tag/v0.14.0) which implements support for `finally`:

---
- ✨ **implementing pipeline level finally** ([tektoncd/pipelines#2661](https://github.com/tektoncd/pipeline/pull/2661))
Users can now specify Tasks within a Pipeline that will always execute, even if Tasks fail, via the new finally clause

- ✨ **adding finally type** (without implementation) at the pipeline level ([tektoncd/pipelines#2650](https://github.com/tektoncd/pipeline/pull/2650))
Introducing a new clause under pipeline named finally where users can specify a list of tasks that will always execute, even if pipeline tasks fail. Adding a new type and its validation here, stay tuned for this functionality. Without implementing the functionality, a list of tasks under finally are validated but not executed at all.

---

This PR proposes to upgrade our supported/required Tekton version from `0.13.0` to `0.14.0`.

Marking this PR as [WIP] while I upgrade my Tekton cluster and run the full set of E2E tests to make sure there are no issues.

/cc @Tomcli @animeshsingh

### DIFF
--- a/sdk/FEATURES.md
+++ b/sdk/FEATURES.md
@@ -1,8 +1,11 @@
-# KFP Tekton compiler features
+# KFP-Tekton Compiler Features
 
-Below are the list of features that are currently available in the KFP Tekton compiler along with its implementation. With the current implementation, we are getting a greater than [80% success rate for compilation tests we are running over approximately 90 pipelines](/sdk/python/tests/README.md) spread across different parts of the Kubeflow Pipeline samples and tests. 
+This document describes the features supported by the _KFP-Tekton_ compiler.
+With the current implementation, the _KFP-Tekton_ compiler is capable of
+[compiling 80%](/sdk/python/tests/README.md) of approximately 90 sample
+and test pipelines found in the KFP repository.
 
-- [Pipeline DSL features with native Tekton implementation](#pipeline-dsl-features-with-native-tekton-implementation)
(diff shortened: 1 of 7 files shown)


ckadner:
I ran the end-to-end tests successfully for the Tekton `0.14.0` release.

The `tkn pipelinerun logs` command does not include the log out for the `finally` task(s) so I updated the "golden" `exit_handler.log`

sdk/FEATURES.md
```diff
@@ -93,7 +96,14 @@ with Tekton's [podTemplate](https://github.com/tektoncd/pipeline/blob/master/doc
 PipelineRun. The [imagepullsecrets](/sdk/python/tests/compiler/testdata/imagepullsecrets.py) python test is an example of how to use this
 feature.
 
-This feature is recently implemented in Tekton and is available on Tekton v0.13.0 onwards.
+This feature has been available since Tekton version `0.13.0`.
+
+### Exit Handler
+
+An exit handler is a component that always executes, irrespective of success or failure, at the end of the pipeline.
```
Tomcli, line 103: Can you add a pointer to the finally test case similar to the rest of the feature sections?

ckadner: I added that and capitalized the headings :-)

PR #222, opened 2020-07-08
Support achive pipelinerun/taskrun logs to S3 storage

### LINKED ISSUE #161: Support log archive to S3 storage.
/kind feature

**Description:**
Argo has support for archiving logs to S3. The UI and cache server then take those information to display the logs or pass it to other artifacts. Since Tekton doesn't support archive logs, we need to have an extra step/sidecar to aggregate logs from main container and sent it to S3.

Design doc summary: https://docs.google.com/document/d/14UF7b5-4YnqssaR2Kkfo0GWSx5uLHF4wHSbyREz73-g/edit

**Additional information:**

fenglixa:
Resolves #161 

Achive pipelinerun/taskrun logs to S3 by fluentd solution. 
https://github.com/eddycharly/dashboard/blob/external-logs/docs/walkthrough/walkthrough-logs.md

User need to install [minio](https://min.io) and [banzaicloud logging operator](https://banzaicloud.com/docs/one-eye/logging-operator/deploy/) under `tools` namespaces before using this feature.

Then user could set `enable_s3_logs=True` and provide parameters of "S3_ACCESSKEY, S3_SECRETKEY and minio_endpoint" to complie the pipelinerun, details see `parallel_join.py` or `compiler_tests.py` file of this PR

/cc @Tomcli

### DIFF
--- a/sdk/FEATURES.md
+++ b/sdk/FEATURES.md
@@ -16,6 +16,7 @@ and test pipelines found in the KFP repository.
     + [Affinity, Node Selector, and Tolerations](#affinity-node-selector-and-tolerations)
     + [ImagePullSecrets](#imagepullsecrets)
     + [Exit Handler](#exit-handler)
+    + [Log Archive](#log-archive)
 - [Pipeline DSL Features with a Custom Tekton Implementation](#pipeline-dsl-features-with-a-custom-tekton-implementation)
   * [Features with the Same Behavior as Argo](#features-with-the-same-behavior-as-argo)
     + [InitContainers](#initcontainers)
@@ -116,6 +117,12 @@ the [exit_handler](/sdk/python/tests/compiler/testdata/exit_handler.py) compiler
 
 The `finally` syntax is supported since Tekton version `0.14.0`.
 
+### Log archive
+Archive tekton pipelinerun logs to S3 storage.
+
+Install [minio](https://min.io) and [banzaicloud logging operator](https://banzaicloud.com/docs/one-eye/logging-operator/deploy/) under `tools` namespaces before using this feature.
+
+Set `enable_s3_logs=True` during complie, for example [parallel join with S3 archive logs](/sdk/python/tests/compiler/testdata/parallel_join.py) compiler test
 
 # Pipeline DSL Features with a Custom Tekton Implementation
 
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -99,6 +99,7 @@ class TektonCompiler(Compiler):
     #
     # Input and output artifacts are hash maps for metadata tracking.
     self.enable_artifacts = None
+    self.enable_s3_logs = None
     self.input_artifacts = {}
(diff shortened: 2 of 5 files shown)


Tomcli:
Thanks @fenglixa, I think we don't need to generate the banzai CRD for every yaml file we generated. It looks like banzai is acted more like a plugin where we can install it as part of our kfp-tekton stack install. And the S3 credentials can be pulled from a config map instead of adding it in the DSL.

Let me do some experiment with the deployment and see how can we make this approach better.

fenglixa:
@Tomcli ,  agree with you, I also feel it's farfetched to integrate such solution to KFP-tekton DSL. It's more like some manual install steps if user want to use this feature. Please let me know if it's OK that our DSL don't need to integrate it. Then this PR could be canceled.

Tomcli:
@fenglixa, I was testing this on our side and we encountered several errors. So creating with the ClusterOutput in this PR we are seeing some key errors for the tags
```
config error file="/fluentd/etc/generated.conf" error_class=Fluent::ConfigError error="chunk_keys specification includes invalid char"
```

We move on by removing the tag field and the config can be installed correctly. However, then we saw the fluentd s3 doesn't support Kube-DNS style endpoint such as minio-service.kubeflow.svc.cluster.local:9000
```
fluent/log.rb:362:error: unexpected error error_class=RuntimeError error="can't call S3 API. Please check your credentials or s3_region configuration. error = #<Seahorse::Client::NetworkingError: Failed to open TCP connection to mlpipeline.minio-service.kubeflow.svc.cluster.local:9000 (getaddrinfo: Name does not resolve)
```

Then we have to use a real cloud object storage from ibm cloud in order to move forward from this. At the end, we still couldn't see the fluentbit sending logs to our object storage. Is the original instruction using minio works for you?

Tomcli:
Also, we discussed with the Watson team and they want to have a more light-weighted option for sending the logs in MVP. One approach @afrittoli suggested is to create a finally task to collect logs at the end of the pipelines and send it to S3. This is probably is the easiest approach and it can be part of the DSL code without adding any infrastructure cost.

fenglixa:
@Tomcli , I haven't fully tested it yet, but the code strictly followed the link from tekton dashboard.
It's meanful in DSL if we use `finally` task to collect pipelinerun logs. Only one uncertain issue I want to find out, where is the pipelinerun logs? I would like to make some research and continue on it.

Tomcli:
@fenglixa An easy way to get the logs is to use the `tkn` CLI https://github.com/tektoncd/cli and run `tkn pipelinerun logs <pipelinerun-name>`. However, this command will get you all the pipeline logs into a single stdout, so we have to parse it and break down the logs into one log file per step. You can get the pipelinerun-name from the pod labels similar to what I did for artifacts. https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/kfp_tekton/compiler/_op_to_template.py#L311

Tomcli:
/close
I will close this for now since we also have the banzai readme. We can open an separate PR for the fluentd approach.

PR #248, opened 2020-07-29
An easy solution of archiving pipeline logs to S3

### LINKED ISSUE #161: Support log archive to S3 storage.

fenglixa:
Resolves #161

Solution: impletmented by `hostPath` of `volumeMount` and` copy_artifacts_step ` in kfp-tekton complier.

**Environment tested:**
```
s3_simply >> tkn pr logs parallel-pipeline -n logging -f
[gcs-download-2 : main] I find thou art no less than fame hath bruited And more than may be gatherd by thy shape Let my presumption not provoke thy wrath
[gcs-download : main] With which he yoketh your rebellious necks Razeth your cities and subverts your towns And in a moment makes them desolate


[gcs-download : copy-artifacts] Added `storage` successfully.
[gcs-download : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_place-scripts-6368a98d9b4571e54b4554e98d4c2a52bd44ff4d7dae7d16b8123cd9ef2317fa.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_place-scripts-6368a98d9b4571e54b4554e98d4c2a52bd44ff4d7dae7d16b8123cd9ef2317fa.log`
[gcs-download-2 : copy-artifacts] Added `storage` successfully.
[gcs-download : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_place-tools-399c93f79305c81d5d18a7b4bb353755094bc5700bdc50ad7dacb15f486b4c46.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_place-tools-399c93f79305c81d5d18a7b4bb353755094bc5700bdc50ad7dacb15f486b4c46.log`
[gcs-download-2 : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_place-scripts-6b949cd25659ea63c539c9becaf22a9b7cb8e29a86315d4861e675a9850d4596.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download-2/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_place-scripts-6b949cd25659ea63c539c9becaf22a9b7cb8e29a86315d4861e675a9850d4596.log`
[gcs-download : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_step-copy-artifacts-3dd191b4bf3d6748e69295ecc9844ec2166be5e87f2ad15547a51fc4b5999d37.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_step-copy-artifacts-3dd191b4bf3d6748e69295ecc9844ec2166be5e87f2ad15547a51fc4b5999d37.log`
[gcs-download : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_step-main-6a7b9350168dca05f5fdfbce074d95365fe464f04df7e1f0357cf650afe4b2b4.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download/parallel-pipeline-gcs-download-ttvf9-pod-8j6l5_logging_step-main-6a7b9350168dca05f5fdfbce074d95365fe464f04df7e1f0357cf650afe4b2b4.log`
[gcs-download : copy-artifacts] Total: 0 B, Transferred: 730 B, Speed: 11.61 KiB/s
[gcs-download-2 : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_place-tools-b79106b1cc1adf9c12f831d9a43d916d768b60efc71ae54e9d66a25b20f47bdf.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download-2/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_place-tools-b79106b1cc1adf9c12f831d9a43d916d768b60efc71ae54e9d66a25b20f47bdf.log`
[gcs-download-2 : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_step-copy-artifacts-e30d84d74964f267c15612537a0fd28e8e3ace8c14f52f7f70ee8357cdbd5b16.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download-2/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_step-copy-artifacts-e30d84d74964f267c15612537a0fd28e8e3ace8c14f52f7f70ee8357cdbd5b16.log`
[gcs-download-2 : copy-artifacts] `/var/log/containers/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_step-main-312e41125b8f958eccd8ec3bcdc5747c57a53015223cac5640d506a0071e14a3.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/gcs-download-2/parallel-pipeline-gcs-download-2-qrdsd-pod-jxhdt_logging_step-main-312e41125b8f958eccd8ec3bcdc5747c57a53015223cac5640d506a0071e14a3.log`

[gcs-download-2 : copy-artifacts] Total: 0 B, Transferred: 1.16 KiB, Speed: 16.98 KiB/s

[echo : main] Text 1: With which he yoketh your rebellious necks Razeth your cities and subverts your towns And in a moment makes them desolate
[echo : main]
[echo : main] Text 2: I find thou art no less than fame hath bruited And more than may be gatherd by thy shape Let my presumption not provoke thy wrath
[echo : main]

[echo : copy-artifacts] Added `storage` successfully.
[echo : copy-artifacts] `/var/log/containers/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_place-scripts-6b58d25daa9602259752c8489625229600f2df59e2198294680b491fb8b6c27d.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/echo/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_place-scripts-6b58d25daa9602259752c8489625229600f2df59e2198294680b491fb8b6c27d.log`
[echo : copy-artifacts] `/var/log/containers/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_place-tools-514e317b8ae727b36971234b89693a14e0d977ee9440fbb7e6a6142e9434e868.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/echo/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_place-tools-514e317b8ae727b36971234b89693a14e0d977ee9440fbb7e6a6142e9434e868.log`
[echo : copy-artifacts] `/var/log/containers/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_step-copy-artifacts-351b110bcbc81b958165d30f30fd0d3dd50de1d381d5b28045fa7a38e1ea2e8b.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/echo/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_step-copy-artifacts-351b110bcbc81b958165d30f30fd0d3dd50de1d381d5b28045fa7a38e1ea2e8b.log`
[echo : copy-artifacts] `/var/log/containers/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_step-main-dd0ab1436bfc8999b4462b5f5f9d20c502ea7c27fa19e9352fbdd1f40da7ced3.log` -> `storage/mlpipeline/artifacts/parallel-pipeline/echo/parallel-pipeline-echo-sf9xm-pod-fjcp7_logging_step-main-dd0ab1436bfc8999b4462b5f5f9d20c502ea7c27fa19e9352fbdd1f40da7ced3.log`
[echo : copy-artifacts] Total: 0 B, Transferred: 1.44 KiB, Speed: 37.35 KiB/s
```
image

image

/cc @animeshsingh @Tomcli


sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -319,7 +319,32 @@ def _process_output_artifacts(outputs_dict: Dict[Text, Any],
...
+                    'name': 'varlibdockercontainers',
+                    'mountPath': '/var/lib/docker/containers',
+                    'readOnly': True
+                }])
+                volume_template.extend([{
+                    'name': 'varlog', 'hostPath': {'path': '/var/log'}
+                }, {
+                    'name': 'varlibdockercontainers', 'hostPath': {'path': '/var/lib/docker/containers'}
+                }])
+            elif artifact['name'] in replaced_param_list:
```
Tomcli, line 347: It should be `if` not `elif`, right now you are removing all the other artifacts when logging is enabled.

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -319,7 +319,32 @@ def _process_output_artifacts(outputs_dict: Dict[Text, Any],
...
+                            'fieldPath': "metadata.labels['tekton.dev/taskRun']"
+                        }
+                    }
+                })
+                volume_mount_step_template.extend([{
+                    'name': 'varlog',
+                    'mountPath': '/var/log'
+                }, {
+                    'name': 'varlibdockercontainers',
+                    'mountPath': '/var/lib/docker/containers',
```
Tomcli, line 339: Current I'm see this when running on IBM Cloud.
```
Added `storage` successfully.
mc: <ERROR> Unable to validate source /var/log/containers/parallel-pipeline-9876e-gcs-download-9b5qx-pod-7xckz_kubeflow_place-scripts-9610f8d694a83791f57386931f55d430775fa3fe3ac2cd4f48182f8e40f51b2d.log
```

This path only has the log symlink for Docker minikube. For IBM Cloud and other cloud provider, they might have different symlink for the log path. Take a look at the example from knative where it covers most of the symlink use case for all cloud providers.
https://github.com/knative/serving/blob/master/config/monitoring/logging/elasticsearch/200-fluentd.yaml#L122-L134

Also, mounting host path for every pipeline could create some security concern. What do you think about converting this part into a daemon set and just use this step to send the PIPELINERUN/PIPELINETASK information to the daemon set?

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -319,7 +319,32 @@ def _process_output_artifacts(outputs_dict: Dict[Text, Any],
         mounted_artifact_paths = []
         for artifact in outputs_dict['artifacts']:
             artifact_name = artifact_to_result_mapping.get(artifact['name'], artifact['name'])
-            if artifact['name'] in replaced_param_list:
+            if artifact['path'] == 'logs_to_S3':
+                copy_artifacts_step['script'] = copy_artifacts_step['script'] + \
+                    'mc cp /var/log/containers/$TASKRUN*$NAMESPACE*.log' + \
```
Tomcli, line 324: Also somehow the ibm cloud cluster doesn't allow copying file directly.
```
mc: <ERROR> Failed to copy `/var/log/containers/parallel-pipeline-c3fae-gcs-download-j7t2j-pod-ll889_kubeflow_step-main-f3a0a7eaa1bb7a4678eb8877777a59cde91d0c8571c81dded25ff60b1e29e7c1.log`. You did not provide the number of bytes specified by the Content-Length HTTP header.
mc: <ERROR> Failed to copy `/var/log/containers/parallel-pipeline-c3fae-gcs-download-j7t2j-pod-ll889_kubeflow_step-copy-artifacts-7e5fe4413051d1984fc1135d1b0fb7300258b6de625fa9aa3b998365213e6354.log`. You did not provide the number of bytes specified by the Content-Length HTTP header.
Total: 0 B, Transferred: 1.05 KiB, Speed: 2 B/s
```

So I have to convert this part into something like
```shell
cat /var/log/containers/$TASKRUN*$NAMESPACE*.log > /tmp/main-log
tar -cvzf main-log.tgz /tmp/main-log
mc cp main-log.tgz storage/mlpipeline/artifacts/$PIPELINERUN/$PIPELINETASK/
```

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: @Tomcli , Thanks for your comments, It should be elif.
because it's under the loop of `for artifact in outputs_dict['artifacts']:`, if it's logging artifact, we should prevent it to continue others.

Tomcli: Thanks, probably some logic modified the artifact list. Let's try to find out where's the real bug.

Tomcli: Oh sorry I think I was running the test with artifacts disabled. It's working now after I turn it back on. You can resolve this comment now.

animeshsingh:
thanks a lot @fenglixa and @Tomcli . Looks exciting.

One comment I had was around shifting the process of adding a "finally" step to server side instead of client? Ideally this is something which an admin of KFP configures, and is enabled/disabled. 

That means then the process of taking Tekton yaml from user, and passing through a function to add "finally" is handled on KFP  server side, before invoking Tekton

Tomcli:
> thanks a lot @fenglixa and @Tomcli . Looks exciting.
> 
> One comment I had was around shifting the process of adding a "finally" step to server side instead of client? Ideally this is something which an admin of KFP configures, and is enabled/disabled.
> 
> That means then the process of taking Tekton yaml from user, and passing through a function to add "finally" is handled on KFP server side, before invoking Tekton

With finally there's still two concerns:
1. Right now we are using Host Path to capture the logs so we can avoid assigning extra RBAC permissions (Read, list, watch pods and pipelinerun) to the pipeline run. This is the common way as we see from fluentd and knative. However, the downside is the client needs to run on the same node, so usually clients are running as daemonset so it can capture logs for all nodes. In this case the client is running inside the same pod on the same node to avoid the daemonset bottleneck.  However as you said this logic is not easy to inject from KFP and users can modify it on their end.

2. If we are going to get the logs using `kubectl logs` or `tkn pipelinerun logs` to avoid the daemonset approach, then we have the permission issue described above because the pod name of each task run is only available during runtime. So users essentially have permission to access all pipeline runs. Another limitation on kubectl is the logs are streamed by kubelet, so we are limited to what kubelet can provide such as file size up to 10MB and log rotation. https://kubernetes.io/docs/concepts/cluster-administration/logging/#logging-at-the-node-level

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: Thanks @Tomcli , add varlibkubeletpods and varlogpods same as knative

sdk/python/kfp_tekton/compiler/_op_to_template.py
fenglixa: I am wonder this is an minio issue, similar issues logged before:
https://github.com/minio/minio/issues/6139
https://github.com/minio/minio/pull/6489

**Anyway, address it in the code here.** 
But unfortunately, my lab cluster has some issues these days, I cannot UT for it right now, will try it later.

fenglixa:
@Tomcli @animeshsingh ,
Thanks for your comments, I don't think we need to consider security issue for the hostPath mounts, why?
1. Most of projects (e.g fluentd and knative) used `hostpath` to get the pod logs officially.
2. I assigned **read permission only** for the volumn mounts.
3. If it's the case, that user modify this yaml file manually will caused security issue, then every tekton yaml file has such risk.  user could modify any tekton yaml file to get the host files. e.g: parallel_join_with_artifacts.yaml

How do you think?

fenglixa:
Addressed the comments and rebase to resolve the conflicts

Tomcli:
Thanks @fenglixa, one last nit. Can you add your test cases into the list of ignored yaml for e2e tests since this will be an experimental feature?
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/compiler_tests_e2e.py#L168

fenglixa:
> Thanks @fenglixa, one last nit. Can you add your test cases into the list of ignored yaml for e2e tests since this will be an experimental feature?
> https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/compiler_tests_e2e.py#L168

Thanks @Tomcli, addressed your comments, please help review. Thanks!

Tomcli:
Thanks @fenglixa

PR #250, opened 2020-07-29
Fix run output crash when artifacts are present

drewbutlerbb4:
Hotfix for a crash coming from the Run Output page when Artifacts are used.


Tomcli:
thanks @drewbutlerbb4 
/lgtm
/approve

PR #259, opened 2020-08-04
fix(sdk): Update SDK to pull artifact related information from annotations

Tomcli:
**Which issue is resolved by this Pull Request:** 
Related #241 

**Description of your changes:**
As we move to multi-user for KFP, we want to inject certain artifact config via the annotations because we cannot change the Tekton API Spec. Thus, update the SDK to pull the related information from annotations without adding any breaking change.

Also we add few extra information to the annotations related to the artifact keys, so the UI can pull them based on similar requirements in #241. 

**Environment tested:**

* Python Version (use `python --version`): 3.7
* Tekton Version (use `tkn version`): 1.14.0
* Kubernetes Version (use `kubectl version`): 1.16
* OS (e.g. from `/etc/os-release`):

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention). 

   PR titles examples:
    * `fix(frontend): fixes empty page. Fixes #1234`
       Use `fix` to indicate that this PR fixes a bug.
    * `feat(backend): configurable service account. Fixes #1234, fixes #1235`
       Use `feat` to indicate that this PR adds a new feature. 
    * `chore: set up changelog generation tools`
       Use `chore` to indicate that this PR makes some changes that users don't need to know.
    * `test: fix CI failure. Part of #1234`
        Use `part of` to indicate that a PR is working on an issue, but shouldn't close the issue when merged.

- [ ] Do you want this pull request (PR) cherry-picked into the current release branch?
    
    If yes, use one of the following options:
  
    * **(Recommended.)** Ask the PR approver to add the `cherrypick-approved` label to this PR. The release manager adds this PR to the release branch in a batch update.
    *  After this PR is merged, create a cherry-pick PR to add these changes to the release branch. (For more information about creating a cherry-pick PR, see the [Kubeflow Pipelines release guide](https://github.com/kubeflow/pipelines/blob/master/RELEASE.md#option--git-cherry-pick).)

Issue #241, opened 2020-07-21
Wanted Tekton feature: manageable artifacts


Tomcli:
/kind feature

**Description:**
Should Tekton have a more mature the concept of artifacts? There are concept of [Tekton artifacts](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#storage-resource) for the pipelineResource and workspaces. However, we don't see any plan for pipelineResource moving to v1beta1 and Tekton artifacts are not manageable on workspaces because they don't have REST api like S3. 

Also in the context of KFP, [KFP artifacts](https://www.kubeflow.org/docs/pipelines/overview/concepts/output-artifact/) need to be stored as tgz and tag with certain metadata. The Tekton pipelineResources/catalog tasks for uploading objects don't have the flexibility for injecting custom pre-processing code (e.g. wrap files into tgz, add metadata JSON file, etc).

We want the Tekton artifacts to able to support:
1. Custom pre-processing code before uploading objects, or options to bundle them into certain format. 
2. Ability to customize any S3 (including minio and non-regional endpoint) or other credentials when uploading those artifact objects no matter which namespace it's in.
3. Annotations/Labels on each task to distinguish output results vs output artifacts, also have annotations for each task to show who owns/generates the input artifacts previously.

Tomcli:
Currently these features are done within our custom task and SDK annotation. Ideally we want these features come out of the box from Tekton

drewbutlerbb4:
Currently, artifacts are not contained in the status of a PipelineRun. In order for the UI to display information about artifacts certain fields must be available describing these artifacts.

The way it is formatted in the Argo version of KFP is that `workflow.status.nodes[x].outputs.artifacts` is a list of artifacts where each artifact has the following information:
```
name: "mlpipeline-ui-metadata"
path: "/tmp/outputs/mlpipeline_ui_metadata/data"
s3:
{
  endpoint: "minio-service.kubeflow:9000"
  bucket: "mlpipeline"
  insecure: true
  accessKeySecret: {name: "mlpipeline-minio-artifact", key: "accesskey"}
  secretKeySecret: {name: "mlpipeline-minio-artifact", key: "secretkey"}
  key: "artifacts/calculation-pipeline-8bhr4/calculation-pipeline-8bhr4-2045269828/mlpipeline-ui-metadata.tgz"
}
optional: true
```
[...]

PR #266, opened 2020-08-10
fix(ui): Update UI to display artifact information

drewbutlerbb4:
Related #241

Fixes the UI to show artifact information and visualizations. More specifically, this PR fixes the Input/Output and Visualizations tabs in for selected nodes in the UI, as well as the Run Output tab in the Run Details page. The information needed to display these improvements is parsed from the SDK changes in #259.

**Environment tested:**

* Python Version (use `python --version`): Python 3.7.7
* Tekton Version (use `tkn version`): Pipeline version: v0.14.2
* Kubernetes Version (use `kubectl version`): Server Version: v1.16.13+IKS

**Checklist:**

   PR titles examples:
    * `fix(frontend): fixes empty page. Fixes #1234`
       Use `fix` to indicate that this PR fixes a bug.
    * `feat(backend): configurable service account. Fixes #1234, fixes #1235`
       Use `feat` to indicate that this PR adds a new feature. 
    * `chore: set up changelog generation tools`
       Use `chore` to indicate that this PR makes some changes that users don't need to know.
    * `test: fix CI failure. Part of #1234`
        Use `part of` to indicate that a PR is working on an issue, but shouldn't close the issue when merged.

    If yes, use one of the following options:
  
    * **(Recommended.)** Ask the PR approver to add the `cherrypick-approved` label to this PR. The release manager adds this PR to the release branch in a batch update.
    *  After this PR is merged, create a cherry-pick PR to add these changes to the release branch. (For more information about creating a cherry-pick PR, see the [Kubeflow Pipelines release guide](https://github.com/kubeflow/pipelines/blob/master/RELEASE.md#option--git-cherry-pick).)


frontend/src/lib/WorkflowParser.ts
```diff
@@ -231,30 +240,64 @@ export default class WorkflowParser {
...
+    })
+
+    outputArtifacts = (rawOutputArtifacts[taskName] || []).map((artifact : any) => {
+      return [ artifact.name, {
+        endpoint: template.endpoint,
+        bucket: template.bucket,
+        insecure: template.insecure,
+        accessKeySecret: template.accessKeySecret,
+        secretKeySecret: template.secretKeySecret,
+        key: artifact.key.replace("$PIPELINERUN", workflow.metadata.name || ""),
```
Tomcli, line 296: Can we add a fallback here when artifact labels are not in the pod, we just ignore and skip the artifact div? Right now it will break if the pod wasn't ready or when it's pointing to a previous version of the pipeline.

frontend/src/lib/WorkflowParser.ts
```diff
@@ -231,30 +240,64 @@ export default class WorkflowParser {
...
 
-    const { inputs, outputs } = workflow.status.nodes[nodeId];
-    if (!!inputs && !!inputs.artifacts) {
-      inputArtifacts = inputs.artifacts.map(({ name, s3 }) => [name, s3]);
-    }
-    if (!!outputs && !!outputs.artifacts) {
-      outputArtifacts = outputs.artifacts.map(({ name, s3 }) => [name, s3]);
+    // Get the task name that corresponds to the nodeId
+    let taskName = "";
+    for (const taskRunId of Object.getOwnPropertyNames(workflow.status.taskRuns)) {
```
Tomcli, line 255: Create a fallback here as well, it will break when the pod/pipeline is not yet running since the taskRun is not yet created.

drewbutlerbb4: Would returning if !workflow.status.taskRuns cover both these cases, since taskRun status is only available when pod/pipeline is running?

Tomcli: As long as it won't break the page it should be good. The ui is constantly pulling new status every 5-10s

drewbutlerbb4:
Thanks @Tomcli, should be fixed now. I fixed some formatting issues as well.

One thing to note is that the output artifacts of a pod will be displayed in the Input/Output selected node tab as soon as the pod is created. If that tab is selected and an output artifact exists then a number of invalid api calls will be made since that output artifact isn't available yet. Since this isn't breaking I have left it as is, but we could instead not show output artifacts until a taskrun is completed if that is preferable.

frontend/src/lib/WorkflowParser.ts
```diff
@@ -231,29 +240,80 @@ export default class WorkflowParser {
...
+    outputArtifacts = (rawOutputArtifacts[taskName] || []).map((artifact: any) => {
+      return [
+        artifact.name,
+        {
+          endpoint: template.endpoint,
+          bucket: template.bucket,
+          insecure: template.insecure,
+          accessKeySecret: template.accessKeySecret,
+          secretKeySecret: template.secretKeySecret,
+          key: artifact.key.replace('$PIPELINERUN', workflow.metadata.name || ''),
```
Tomcli, line 312: For backward compatible purpose, can you check is the artifact key exists? If it's not exist, then we don't display the artifact section.

Tomcli:
> Thanks @Tomcli, should be fixed now. I fixed some formatting issues as well.
> 
> One thing to note is that the output artifacts of a pod will be displayed in the Input/Output selected node tab as soon as the pod is created. If that tab is selected and an output artifact exists then a number of invalid api calls will be made since that output artifact isn't available yet. Since this isn't breaking I have left it as is, but we could instead not show output artifacts until a taskrun is completed if that is preferable.

If it's an easy change then we can let the artifact wait until the task is completed. This is mainly a limitation from tekton because tekton doesn't have anyway to inject annotations during runtime, so we have prepopulate the artifact relationship which cause this race condition issue.

drewbutlerbb4:
@Tomcli I have added the backwards compatibility check and I only allowed output artifacts to be shown if they were in the COMPLETED or SUCCEEDED phase for now.

Tomcli:
Thanks @drewbutlerbb4 
/lgtm
/approve

Issue #277, opened 2020-08-19
Refactor kfp-tekton artifact implementation


Tomcli:
/kind feature

**Description:**
Items:
- [ ] Configurable annotations for injecting default credentials (#278)
- [ ] Expose artifact mount dependencies to annotation. Move client side logic to the server side and inject the necessary post-processing steps/

PR #278, opened 2020-08-19
Feature: Configurable artifact credentials and annotations

### LINKED ISSUE #269: Create dedicated configmap for kfp-tekton configuration
/kind feature

**Description:**
In kfp-tekton, we introduced many custom KFP concepts that aren't available in the native tekton engine. (e.g. artifacts, loggings, etc). Currently many of these configurations are pulling/reusing from the Argo configmap so we can have the consistent behaviors. However, we should now create and maintain our own version of configmap unique to kfp-tekton so that we can opt-out Argo in our deployment while introducing new configmap fields.

In addition, we want to move this configmap as a dependency for the kfp-tekton api deployment so that the client won't have control of the S3 and logging configuration used in KFP.

Tomcli:
Resolves #269 
related: #277 

Implement our own configmap for things that are unique to kfp-tekton. This includes configurable artifact credentials on the server side (since Tekton doesn't support it) and other flags/settings that we want to add later on.  This PR builds the foundation for the artifact refactoring where we will be moving all the SDK client side logic to the server side.

**Environment tested:**

* Python Version (use `python --version`): 3.7
* Tekton Version (use `tkn version`): 0.15

### DIFF
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -84,6 +84,7 @@ TEKTON_TASKRUN_LABEL_KEY = 'tekton.dev/taskRun'
 TEKTON_PIPELINETASK_LABEL_KEY = 'tekton.dev/pipelineTask'
 TEKTON_INPUT_ARTIFACT_ANNOTATION_KEY = 'tekton.dev/input_artifacts'
 TEKTON_OUTPUT_ARTIFACT_ANNOTATION_KEY = 'tekton.dev/output_artifacts'
+TEKTON_BUCKET_ARTIFACT_ANNOTATION_KEY = 'tekton.dev/artifact_bucket'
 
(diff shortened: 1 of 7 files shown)


animeshsingh:
thanks @Tomcli 

What are the cons of using Argo configmap? Also not sure what this statement means

_In addition, we want to move this configmap as a dependency for the kfp-tekton api deployment so that the client won't have control of the S3 and logging configuration used in KFP._

cc @fenglixa - can you review this as well

Tomcli:
> What are the cons of using Argo configmap? Also not sure what this statement means
>

Right now the Argo configmap is part of the Argo component under Kubeflow/manifests, so reusing the Argo configmap will require us to install Argo as part of our Tekton deployment.

> _In addition, we want to move this configmap as a dependency for the kfp-tekton api deployment so that the client won't have control of the S3 and logging configuration used in KFP._

This configmap is one of the key dependency for PR #281

PR #285, opened 2020-08-26
update kfp tekton install steps

animeshsingh:
update installation instructions to point to openshift and simplify installation flow


Tomcli:
Thanks @animeshsingh, i will do a follow up pr to create the table of contents and trim out some ibm reference and hyperlink the cloud specific instructions.

PR #290, opened 2020-08-27
Make should use specified VENV

ckadner:

Update `.travis.yml` and `Makefile` to avoid duplicate/redundant `pip install steps` by making sure that the `make` targets use the already existing Python `virtualenv` created by Travis, instead of creating a second `VENV` inside the `make` process. This should speed up Travis build times and reduce resource consumption slightly, but at the very least it makes for more readable build logs.

**Environment tested:**

* Python Version (use `python --version`): `3.5`, `3.6`, `3.7`
* Tekton Version (use `tkn version`): n/a
* Kubernetes Version (use `kubectl version`): n/a
* OS (e.g. from `/etc/os-release`): Mac OS Catalina, Linux Ubuntu 16.04.6 LTS

/cc @Tomcli

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,12 @@ help: ## Display the Make targets
 .PHONY: venv
 venv: $(VENV)/bin/activate ## Create and activate virtual environment
 $(VENV)/bin/activate: sdk/python/setup.py
+# create/update the VENV when there was a change to setup.py
+# check if kfp-tekton is already installed (Travis/CI did during install step)
+# use pip from the specified VENV as opposed to any pip available in the shell
 	@echo "VENV=$(VENV)"
 	@test -d $(VENV) || python3 -m venv $(VENV)
-	pip install -e sdk/python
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ venv: $(VENV)/bin/activate ## Create and activate virtual environment
 $(VENV)/bin/activate: sdk/python/setup.py
 	@echo "VENV=$(VENV)"
 	@test -d $(VENV) || python3 -m venv $(VENV)
-	pip install -e sdk/python
+	@$(VENV)/bin/pip show kfp-tekton >/dev/null 2>&1 || $(VENV)/bin/pip install -e sdk/python
```
ckadner, line 32: - check if `kfp-tekton` is already installed (in Travis/CI context it will have been done during `install`)
- use the specified `VENV` as opposed to any `pip` availlable by the shell

animeshsingh: Maybe add the above text in comments?

Tomcli: @ckadner do you want to add the above comments in the Makefile?

PR #299, opened 2020-09-04
feat(backend): new server API to read run log. Fixes #298.

### LINKED ISSUE #298: Server API to read run log
/kind feature

**Description:**
Run log is currently retrieved directly from Pod or from artifact location in the BFF endpoint. There is no KFP server API that allows for that, no way to expose this capability in SDK/CLI. Additionally, from the KFP adaptation perspective it creates a need to duplicate code.

**Additional information:**
It is much natural to introduce the new KFP server API endpoint that allows to read run log in the similar way as current implementation BFF endpoint. Additionally, proper modularisation should allow to make log archive logic customisable.

rafalbigaj:
Resolves #298

The new backend API to read run log is proposed in place of the existing BFF endpoint.

New endpoint: `GET /apis/v1beta1/runs/{run_id}/nodes/{node_id}/log`

**Environment tested:**

* Python Version (use `python --version`): Python 3.7.7
* Tekton Version (use `tkn version`): v0.15.0
* Kubernetes Version (use `kubectl version`): v1.17.11+IKS
* OS (e.g. from `/etc/os-release`): Debian GNU/Linux 10 (buster)

**Checklist:**

   PR titles examples:
    * `fix(frontend): fixes empty page. Fixes #1234`
       Use `fix` to indicate that this PR fixes a bug.
    * `feat(backend): configurable service account. Fixes #1234, fixes #1235`
       Use `feat` to indicate that this PR adds a new feature. 
    * `chore: set up changelog generation tools`
       Use `chore` to indicate that this PR makes some changes that users don't need to know.
    * `test: fix CI failure. Part of #1234`
        Use `part of` to indicate that a PR is working on an issue, but shouldn't close the issue when merged.

    If yes, use one of the following options:
  
    * **(Recommended.)** Ask the PR approver to add the `cherrypick-approved` label to this PR. The release manager adds this PR to the release branch in a batch update.
    *  After this PR is merged, create a cherry-pick PR to add these changes to the release branch. (For more information about creating a cherry-pick PR, see the [Kubeflow Pipelines release guide](https://github.com/kubeflow/pipelines/blob/master/RELEASE.md#option--git-cherry-pick).)


backend/src/apiserver/archive/log.go
```diff
@@ -0,0 +1,169 @@
...
+// CopyLogFromArchive copies a task run archived log into expected format.
+func (a *LogArchive) CopyLogFromArchive(logContent []byte, dst io.Writer, opts ExtractLogOptions) error {
+	reader, err := decompressLogArchive(logContent)
+	if err != nil {
+		return err
+	}
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		// line := strings.Trim(scanner.LogFormatText(), "\n\r\t ")
```
Tomcli, line 90: Is this commented code for debugging?

backend/src/apiserver/archive/log.go
```diff
@@ -0,0 +1,169 @@
...
+	if taskRun == nil {
+		err = util.NewBadRequestError(errors.New("archived log cannot be read"),
+			"Failed to retrieve the task run from the runtime workflow")
+		return
+	}
+
+	workflowName := workflow.Name
+	taskName := taskRun.PipelineTaskName
+
+	key = fmt.Sprintf("/artifacts/%s/%s/%s--step-main_log.gz", workflowName, taskName, taskRunId)
```
Tomcli, line 76: FYI we are not using taskRunId as our artifact path because we cannot determine this value for the input artifacts during static time.

Tomcli:
Thanks @rafalbigaj, I will also let few of our team members to take a look and also see what KFP's view point on this.

backend/src/apiserver/archive/log.go
rafalbigaj: @Tomcli what do you use for now in the log file name?

backend/src/apiserver/archive/log.go
rafalbigaj: Thanks, that's redundant comment. I've already removed it.

backend/src/apiserver/archive/log.go
Tomcli: right now we have `/artifacts/workflowName/taskName/main-log.tgz` as default, but you can update the path in the configmap over [here](https://github.com/kubeflow/kfp-tekton/blob/1296cb1c7f86237cd276f7fabd1500e69527f907/manifests/kustomize/env/platform-agnostic/kfp-pipeline-config.yaml#L16-L23)

rafalbigaj:
@Tomcli do we need feedback from others to proceed with this PR?

animeshsingh:
thanks @rafalbigaj 
few thoughts I had on this, and want to get your inputs
1. you have any thoughts vis a vis performance aspects on real time log streaming if going through APIs?
2. Currently KFP uses BFF, and should we do it in a more generic way so that it can work with both Argo/Tekton and can also be PRed back to mainstream KFP?
2. If not, can it done based on boolean logic, so that the "platform" deployer can choose to enable APIs to be used for log-streaming vs BFF?

Tomcli:
@rafalbigaj You also need to update the rbac for the api-server. The default rbac won't let you log any pod. 
https://github.com/kubeflow/kfp-tekton/blob/f2b33cecf79fb69b33b604be9dfcb789831fb3c1/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-role.yaml#L10-L11

So add `- pods/log` under pods

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -544,6 +550,70 @@ func (r *ResourceManager) RetryRun(runId string) error {
...
 }
 
+func (r *ResourceManager) ReadLog(runId string, nodeId string, dst io.Writer) error {
+	run, err := r.checkRunExist(runId)
+	if err != nil {
+		return util.NewBadRequestError(errors.New("log cannot be read"), "Run does not exist")
+	}
+
+	err = r.readRunLogFromPod(run, nodeId, dst)
+	if err != nil {
```
Tomcli, line 560: For this error, can we log it with info or warning? On my env, I was having a rbac permission error, but I couldn't see it from the logs since it get ignored/overwritten over here.

Tomcli:
> @Tomcli do we need feedback from others to proceed with this PR?

From https://github.com/kubeflow/pipelines/issues/4468, the KFP team is welcome this API and our team has no objection. So we can proceed PR and hopefully this can merge back to KFP as well.

tonanhngo:
Would this feature work with Argo as well?  It seems to be a generally useful feature.  If so, maybe we should create the PR in KFP in the first place to have better adoption and reduce the effort in merging later.
kubectl can also stream log, something like:
kubectl logs -f ...
Would this meet the requirement, at least for the cli?

Tomcli:
> Would this feature work with Argo as well? It seems to be a generally useful feature. If so, maybe we should create the PR in KFP in the first place to have better adoption and reduce the effort in merging later.
> kubectl can also stream log, something like:
> kubectl logs -f ...
> Would this meet the requirement, at least for the cli?

I agree we should also push this feature to KFP, but the code as it won't work on Argo since there are some dependency conflicts with the Argo and Tekton client. This is one of the reasons I removed all the Argo code for now until this issue https://github.com/kubeflow/pipelines/issues/3770 is resolved. KFP team might remove Bazel soon so hopefully we can have an easier time to update Argo in the near future.

As for this PR, we can get it into our kfp-tekton first since we don't know how much time it will take to merge into KFP.

animeshsingh:
@Tomcli do we have this working with UI and SDK? What's the exposure? Is there any lag/limits et all on the amount of logs to be streamed through API vs BFF?

Tomcli:
> @Tomcli do we have this working with UI and SDK? What's the exposure? Is there any lag/limits et all on the amount of logs to be streamed through API vs BFF?
After fixing it with the above rbac, this PR will work with the UI. Since this is a new concept for SDK, we will need to implement a new python client function to handle this new API. In theory it should work because the UI is using the same API endpoints and schema.

As for the lag and limits, I can create some stress tests and see can it break. But the performance may vary depending on the real-time cloud environment since we are using shared vm for IKS.

Tomcli:
In terms of exposure is the same as BFF since users already can get the logs via curl before, but the api syntax in this PR is bit more straight forward here and in theory should perform better in golang.

Tomcli:
So performance wise I created the pipeline below for testing. Overall the logging from API is doing better than in BFF since golang is more efficient than nodejs at streaming logs. The only con I can see here is if someone abuse this API and streaming a bun of logs, it will slow down the whole apiserver instead of the BFF pod. 

I would say if we really want to make it right, then we should create a new logging microservice (such as fluentd with elastic search) for just streaming logs in real-time. But since we don't want to increase the infrastructure cost, I suggest we should have some flags to let the admin decide whether to stream logs from BFF or the apiserver.

```python
from kfp import dsl
import kfp.components as comp
def gen_log(loop_number: int) -> str:
   '''Generate logs'''
   for i in range(loop_number):
       print("This is loop #%s" % str(i))
   return "done"

gen_log_op = comp.func_to_container_op(gen_log)

@dsl.pipeline(
    name='gen_log',
    description='gen_log pipeline'
)
def gen_log_pipeline(
    loop_number="1000000000",
):
    gen_log_comp = gen_log_op(loop_number)

if __name__ == '__main__':
    TektonCompiler().compile(gen_log_pipeline, 'test_pipeline.yaml')
```

tonanhngo:
I would agree that if the goal is to have a full-function, scaleable logging support for production use, then there are good existing solutions.  In this case, it doesn't seem some additional infrastructure should be a concern.  In our environment, @tedhtchang has set up kubeflow with fluentd+forwarder+elasticsearch+kibana and it works very well.

rafalbigaj:
@Tomcli @animeshsingh I have prepared a PR to the mainstream KFP repository: https://github.com/kubeflow/pipelines/pull/4493.

rafalbigaj:
@Tomcli @tonanhngo thank you for all your comments and suggestions!

backend/src/apiserver/archive/log.go
rafalbigaj: I have added server configuration as well. Thanks

backend/src/apiserver/resource/resource_manager.go
rafalbigaj: Error handling is enhanced now

rafalbigaj:
@Tomcli rbac role is now updated.

rafalbigaj:
@Tomcli @animeshsingh I think, there is no open question / comment.
How could we proceed with this PR?
Should we wait for review of https://github.com/kubeflow/pipelines/pull/4493 or could this one go first?

backend/src/apiserver/config/config.json
```diff
@@ -11,6 +11,10 @@
     "BucketName": "mlpipeline",
     "PipelinePath": "pipelines"
   },
+  "ArchiveConfig": {
+    "LogFileName": "step-main.log",
```
Tomcli, line 15: Thanks @rafalbigaj, it looks great. Our current logs are pushed as `step-main.tgz`. If you think `.log` is better, we can update our default archive logs function to push it without compression.

rafalbigaj: @Tomcli I have updated the default. Thanks!

Tomcli:
/lgtm
Thanks @rafalbigaj 
We probably need a follow up PR to expose these flags to the kfp-tekton configmap, but this PR should be good to go.

animeshsingh (approved):
/lgtm

rafalbigaj:
@animeshsingh thanks!

PR #370, opened 2020-11-24
add reference to watson blog post

animeshsingh:


Tomcli:
/lgtm

FYI, opening new branches in https://github.com/kubeflow/kfp-tekton could make cloning the whole repo bigger than usual. We should clean up branches that are no longer necessary (e.g. animeshsingh-patch-1, animeshsingh-patch-2, etc)

Issue #413, opened 2021-01-11
DSL enhancement to support loops feature for kfp-tekton


fenglixa:
/kind feature

**Description:**
[A clear and concise description of what your proposal. What problem does it solve?]
The PR of pipeline loop I already provided by #412
Log this issue to trace the DSL enhancement of supporting loops feature for kfp-tekton, includes both static and dynamic loop.
The loop defination will be as below:
```
    op0 = dsl.ContainerOp(
        name="my-out-cop0",
        image='python:alpine3.6',
        command=["sh", "-c"],
        arguments=[
            'python -c "import json; import sys; json.dump([i for i in range(20, 31)], open(\'/tmp/out.json\', \'w\'))"'],
        file_outputs={'out': '/tmp/out.json'},
    )

    with dsl.ParallelFor(op0.output) as item:
        op1 = dsl.ContainerOp(
            name="my-in-cop1",
            image="library/bash:4.4.23",
            command=["sh", "-c"],
            arguments=["echo do output op1 item: %s" % item],
        )
```

OR:

```
    with dsl.ParallelFor({“from”: 1, “to”: 4, “step”: 1}) as item:
        op1 = dsl.ContainerOp(
            name="my-in-cop1",
            image="library/bash:4.4.23",
            command=["sh", "-c"],
            arguments=["echo do output op1 item: %s" % item],
        )

```
**Additional information:**
[Miscellaneous information that will assist in solving the issue.]
/assign

The PR will be provided later after my fully UTed

PR #440, opened 2021-01-28
feat(backend): Support dsl to define the pipeline loops via tekton

fenglixa:
This PR resoved part of #413. for the basic defination of the loops. support both static and dynamic loop.

PipelineLoops is a feature for running the same component tasks multiple times. right now Tekton support the loop pipeline/tasks via custom tasks. The controller we named as "PipelineLoop".

The DSL defination like below as example:
```
loop_args = [1, 2, 3]
    with dsl.ParallelFor(loop_args) as item:
        op1 = dsl.ContainerOp(
            name="my-in-coop1",
            image="library/bash:4.4.23",
            command=["sh", "-c"],
            arguments=["echo op1 %s %s" % (item, my_pipe_param)],
        )

        op2 = dsl.ContainerOp(
            name="my-in-coop2",
            image="library/bash:4.4.23",
            command=["sh", "-c"],
            arguments=["echo op2 %s" % item],
        )
```

/cc @Tomcli 

**Environment tested:**

**Checklist:**

- [x] Do you want this pull request (PR) cherry-picked into the current release branch?


Tomcli:
Thanks @fenglixa, I think we need to find a way to put these into one pipelinerun yaml. Right now the new implementation will break our test since it expects the pipelineloop need to be deployed before pipelinerun.

sdk/python/tests/compiler/testdata/loop_static.py
```diff
@@ -17,20 +17,20 @@
 
 @dsl.pipeline(name='my-loop-pipeline')
 def pipeline(my_pipe_param='10'):
-    loop_args = [{'A_a': 1, 'B_b': 2}, {'A_a': 10, 'B_b': 20}]
```
Tomcli, line 20: If the custom task cannot support looping over a dictionary, should we keep the old implementation for flatten list of dictionary arguments?

fenglixa: The dictionary supporting is on the way, I am doing it right now, expect it will be ready soon. To support it, I need to update the reconciler of pipelineloop, since this is not a dataStage case, that's why I didn't impletmented it.

fenglixa:
> Thanks @fenglixa, I think we need to find a way to put these into one pipelinerun yaml. Right now the new implementation will break our test since it expects the pipelineloop need to be deployed before pipelinerun.

I can put the pipelineloop cr yaml files with the pipelinerun yaml into one yaml file, and enhance 'compiler_tests.py' to only validate the pipelinerun object. This way will workaround the testing. How do you think?

Tomcli:
For the test, since we don't have the exception to block dynamic parameters, we will have to regenerate the test report.
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/test_kfp_samples.sh

fenglixa:
> For the test, since we don't have the exception to block dynamic parameters, we will have to regenerate the test report.
> https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/test_kfp_samples.sh

Yeah, that's the case, seems we don't need extral actions to make the test works.

Tomcli:
Since this is still experimental, we will cut our kfp-tekton 0.6 release tomorrow or monday so our other users won't break their current implementation. I will merge this PR once the 0.6 release is cut.

By the way, having you try to use the dynamic parameters with the master branch of Tekton? I think they added the support for custom task results.

/lgtm
/hold

fenglixa:
> Since this is still experimental, we will cut our kfp-tekton 0.6 release tomorrow or monday so our other users won't break their current implementation. I will merge this PR once the 0.6 release is cut.
> 
> By the way, have you try to use the dynamic parameters with the master branch of Tekton? I think they added the support for custom task results.
> 
> /lgtm
> /hold

Thanks @Tomcli I didn't get the chance to try Tekton with master branch. I can do it later before next pr (support dictionary)

PR #494, opened 2021-03-11
Fix the copy result issue when it's already tekton result path

### LINKED ISSUE #493: Copy results failed when file output already specify /tekton/result path
/kind bug
Here's a very simple test program:
```
import kfp
from kfp import dsl


def random_num_op(low, high):
    """Generate a random number between low and high."""
    return dsl.ContainerOp(
        name='Generate random number',
        image='python:alpine3.6',
        command=['sh', '-c'],
        arguments=['python -c "import random; print(random.randint($0, $1))" | tee $2', str(low), str(high), ' /tmp/OutPut'],
        file_outputs={'OutPut': '/tekton/results/output'}
    )


def print_op(msg):
    """Print a message."""
    return dsl.ContainerOp(
        name='Print',
        image='alpine:3.6',
        command=['echo', msg],
    )


@dsl.pipeline(
    name='Conditional execution pipeline',
    description='Shows how to use dsl.Condition().'
)
```

Notice the file output is already tekton result path: ```file_outputs={'OutPut': '/tekton/results/output'}```,  but after compile still copy results is generated:

```
        - image: busybox
          name: copy-results
          script: |
            #!/bin/sh
            set -exo pipefail
            cp  /tekton/results/output $(results.output.path);
```

Should not add the copy steps when it's already in the correct path

pugangxa:
Resolves #493 

Just add guard code

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_op_to_template.py
+++ b/sdk/python/kfp_tekton/compiler/_op_to_template.py
@@ -246,6 +246,8 @@ def _process_parameters(processed_op: BaseOp,
                             need_copy_step = False
                         args.append(a)
                     s['args'] = args
+                if path == '/tekton/results/' + sanitize_k8s_name(name):
+                    need_copy_step = False
             # If file output path cannot be found/replaced, use emptyDir to copy it to the tekton/results path
             if need_copy_step:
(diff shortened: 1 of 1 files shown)


fenglixa:
Thanks @pugangxa 
/lgtm
/approve

PR #499, opened 2021-03-16
Enhance AnySequencer to make it an op

pugangxa:
Enhance AnySequencer to make it an container op with name which is easier to reuse

Because Any Sequencer is created implicitly, by applying after_any function to another op. There are two problems here:

- any-sequencers can get no name or other user-defined metadata
- each op needs its own any-sequencer, even if it's identical to another one

I'd suggest to make it an op instead (dsl.AfterAny? dsl.AnySequencer?), and make after_any a simple utility function which creates a new after-any op and sets a dependency on it to the argument op.


sdk/python/tests/compiler/testdata/any_sequencer.py
```diff
@@ -49,8 +48,7 @@ def any_sequence_pipeline(
         image="registry.access.redhat.com/ubi8/ubi-minimal",
         command=["/bin/bash", "-c"],
         arguments=["sleep 30"]
-    ).apply(after_any([task1, task2, task3]))
-
+    ).apply(after_any([task1, task2, task3], "any_test"))
```
fenglixa, line 52: Thanks Gang, could you let me know what's the mean of "any_test" defined here?

fenglixa: Any sequency task name?

pugangxa: Yes, this is the name. See the context:
```
Because Any Sequencer is created implicitly, by applying after_any function to another op. There are two problems here:

**any-sequencers can get no name or other user-defined metadata**
each op needs its own any-sequencer, even if it's identical to another one
```

fenglixa:
Thanks @gang, For the PR. the CI test failed, seems some `license header` was not added caused it. 
You can run `make lint` before commit to check what's the problem

sdk/python/kfp_tekton/dsl/_any_seqencer.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2019 Google LLC
```
fenglixa, line 1: 2019 -> 2021

pugangxa:
And it can be easily reused if there's multiple tasks depend on the any sequencer, for example task1 and task2 all depend on the AnySequencer:
```
seq = AnySequencer(
  [op0, op1],
  # naming the Any Sequencer, like any other op
  name="wait-for-any-file"
)
task1 = dsl.ContainerOp(
......
).after(seq)
task2 = dsl.ContainerOp(
......
).after(seq)
```

Tomcli:
before merging this PR, just make sure to run the following commands to regenerate the tests and fix the python lint
```
make unit_test GENERATE_GOLDEN_YAML=True
make lint
make check_license
```

pugangxa:
> before merging this PR, just make sure to run the following commands to regenerate the tests and fix the python lint
> 
> ```
> make unit_test GENERATE_GOLDEN_YAML=True
> make lint
> make check_license
> ```

Thanks Tommy, got it!

fenglixa:
Thanks @pugangxa 
/lgtm

jinchihe:
@pugangxa Looks great, thanks Gang.

PR #507, opened 2021-03-19
Add automated labels and annotations for pipeline caching

### LINKED ISSUE #497: automate labels and annotations for pipeline cache
/kind feature

**Description:**
We has done the implements for `cache server` to meet the `tekton` for backend. and try to enable the function in `DSL` side.

Basically, to enable the `cache` feature just need add several specific `label`s to the sub task.
By review the original implements(Argo), it add the label in the api level(a global setting), it's not work in `tekton`, in `tekton` scenario, we expect enable `cache` in sub tasks, since the `output cache` is based on `Result`, not every task has result.

For now, we just let user add the label manually in DSL code, like this:
```
    task1.add_pod_label('pipelines.kubeflow.org/cache_enabled', "true")
    task1.add_pod_annotation('tekton.dev/template', "")
```

But I think we can make it more grace to wrapper/abstract it, maybe a parameter: `enalbe-cache` and compiler will add these `label`s automatically。

**Additional information:**

drewbutlerbb4:
Resolves #497 

Add caching by default unless a flag `--disable-cache` is added.

**Environment tested:**

**Checklist:**


drewbutlerbb4:
/cc @Tomcli 
Should the testdata pipelineruns have caching by default? If so I will update all the pipelines, otherwise I can add a `--disable-cache` flag when compiling them.

Tomcli:
@drewbutlerbb4 you can add the cache flags to all the testdata, we can disable the cache if needed by removing the cache server.

Tomcli:
also under the feature page, can you add a section on what annotations are for the cache and how to disable it?
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/FEATURES.md

Thanks

sdk/FEATURES.md
```diff
@@ -22,6 +22,7 @@ and test pipelines found in the KFP repository.
     - [Any Sequencer](#any-sequencer)
     - [Tekton Pipeline Variables](#tekton-pipeline-variables)
     - [Sidecars](#sidecars)
+    - [Caching](#caching)
```
Tomcli, line 25: KFP caching is not a native Tekton implementation, can you move this to the `Pipeline DSL Features with a Custom Tekton Implementation` section? thanks.

Tomcli:
Thanks @drewbutlerbb4 
@vincent-pli does it satisfy your requirement? we can make this as default for all pipelines.

vincent-pli:
@drewbutlerbb4 @Tomcli 
I checked the code, sees the `enable-cache` flag is for `pipelinerun` level, is that means if the flag is true, every sub tasks in the `pipeline` will be cache enabled?

If that true, it cannot satisfy the requirements, it required the `enable-cache` flag is `task` level, I means each task could disable/enable cache.

I'm not oppose to have a `pipeline` level flag, but the `task` level flag is also required and if there is conflict between the two flags, chose the `task` level one. thanks

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -760,6 +766,7 @@ def compile(self,
               pipeline_func,
               package_path,
               type_check=True,
+              enable_cache=True,
```
Tomcli, line 769: If we are going with annotations as flag, maybe we don't need to have this argument over here

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -424,12 +424,18 @@ def _create_pipeline_workflow(self, args, pipeline, op_transformers=None, pipeli
...
-        if template['metadata'].get('labels', None):
-          task_ref['taskSpec']['metadata'] = task_ref['taskSpec'].get('metadata', {})
-          task_ref['taskSpec']['metadata']['labels'] = template['metadata']['labels']
-        if template['metadata'].get('annotations', None):
-          task_ref['taskSpec']['metadata'] = task_ref['taskSpec'].get('metadata', {})
-          task_ref['taskSpec']['metadata']['annotations'] = template['metadata']['annotations']
+        task_metadata = task_ref['taskSpec'].get('metadata', {})
+        task_ref['taskSpec']['metadata'] = task_metadata
+        task_metadata['labels'] = template['metadata'].get('labels', {})
+        task_metadata['labels']['pipelines.kubeflow.org/pipelinename'] = task_metadata['labels'].get('pipelines.kubeflow.org/pipelinename', '')
```
Tomcli, line 430: Not sure will the pipeline level labels will overwrite the task one, but make sure to check the pipeline level labels in this logic as well

Tomcli: Pipeline level labels are injected over here https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/kfp_tekton/compiler/compiler.py#L568-L574

sdk/python/kfp_tekton/compiler/compiler.py
drewbutlerbb4: @Tomcli 
I left this in so caching would be enabled by default, but removing caching from the whole pipeline could be easily done by adding a single flag `--disable-cache` at compile. If we remove this then someone who wants to disable caching would have to add `task1.add_pod_label('pipelines.kubeflow.org/cache_enabled', "true")` to all of their tasks.
If you are saying we should remove enabling caching by default then I can remove all that logic?

Tomcli: you can take a look at this example, we added a new way to have annotations shared for the whole pipeline. If someone wants to disable it, they can do it using the tekton_pipeline_conf
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/testdata/tekton_pipeline_conf.py

The original Kubeflow way is to add annotations to each task as well, but just in Argo there's no annotations applied to the whole pipeline because it's not supported.

drewbutlerbb4:
@vincent-pli @Tomcli 
Thanks for your comments
@vincent-pli 
I have changed it so that caching can be enabled on the task level by adding a single label as below. Or disabled by setting to `false`.
``` task1.add_pod_label('pipelines.kubeflow.org/cache_enabled', "true")```

In the special case that a `pipelines.kubeflow.org/cache_enabled` label is set on the pipeline level as well as the task level, the Tekton default is to overwrite the task level with the pipeline level version. So if a pipeline label `pipelines.kubeflow.org/cache_enabled` is set then all task labels of `pipelines.kubeflow.org/cache_enabled` will be ignored. Would you prefer that task labels override pipeline labels?

File: sdk/python/kfp_tekton/compiler/main.py [hunk 4]
```diff
@@ -49,7 +49,7 @@ def parse_arguments():
   return args
 
 
-def _compile_pipeline_function(pipeline_funcs, function_name, output_path, type_check):
+def _compile_pipeline_function(pipeline_funcs, function_name, output_path, type_check:
```
Tomcli, line 52: typo that fails the tests

drewbutlerbb4: thanks!

Tomcli: can you also fix the python lint? you can check the errors with `make lint`

vincent-pli:
@drewbutlerbb4 
Yes, but user can only set `task` level labels.
BTW, could we add a sample for it?

drewbutlerbb4:
Thanks for your comments @Tomcli @vincent-pli 
Task level labels overwrite pipeline level labels now and I added a sample which shows the results of caching enabled and caching disabled.

sdk/FEATURES.md
```diff
@@ -234,6 +235,12 @@ It also includes several annontations, `tekton.dev/input_artifacts` and `tekton.
 The current implementation is relying on the existing KFP's minio setup for getting the default credentials. These default credentials can be updated
 via a Kubernetes configmap.
 
+### Caching
+
+By default compiling a pipeline will add metadata annotations and labels so that results from tasks within a pipeline run can be re-used if that task is reused in a new pipeline run. This saves the pipeline run from re-executing the task when the results are already known. 
+
+The specific annotations and labels that are added to the task spec metadata to enable caching are: `annotations={'tekton.dev/template': ""}` and `labels={'pipelines.kubeflow.org/cache_enabled': 'true', 'pipelines.kubeflow.org/pipelinename': '', 'pipelines.kubeflow.org/generation': ''}`.
+
```
Tomcli, line 243: Can you tell the users that cache is enabled by default. If they want to disable cache, they need to use the label from your example.
```
task2.add_pod_label('pipelines.kubeflow.org/cache_enabled', 'false')
```

Tomcli:
/lgtm
/hold
@vincent-pli any last comment?
@drewbutlerbb4 There's a broken kubeflow/website link in the test, I will update it later since they just finished the refactoring.

Tomcli:
/lgtm
/unhold

@vincent-pli let us know if there's other default behavior you want to add to the SDK. Right now we will enable the cache by default and this can be disabled with a cache label.

PR #489, opened 2021-03-08
fix(frontend): Display loop UI for pipelines page

fenglixa:
**Which issue is resolved by this Pull Request:** 
Resolves #

**Description of your changes:**
Address the loops UI dispaly in static pipelines page. 
Run page was blocked by issue #488 now.

**Environment tested:**
image

* Python Version (use `python --version`):
* Tekton Version (use `tkn version`):
* Kubernetes Version (use `kubectl version`):
* OS (e.g. from `/etc/os-release`):

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention). 

- [x] Do you want this pull request (PR) cherry-picked into the current release branch?

    [Learn more about cherry-picking updates into the release branch](https://github.com/kubeflow/pipelines/blob/master/RELEASE.md#cherry-picking-pull-requests-to-release-branch).

PR #512, opened 2021-03-25
feature(backend): Allow KFP apiserver to create recursion pipelineLoop CRs for the users

### LINKED ISSUE #496: Add support for auto apply custom task resources until taskSpec is implemented.
/kind feature

**Description:**
Given the slow pace on the taskSpec support for custom task, we want to have better optimization and user experience for the time being. Thus, we propose the following solution.

1. Add custom task CR information in the annotations, so the information can be carried to the API and UI without any API schema changes.
2. Add flags in API to support auto CR creations. Give a list of CR specs in the annotations, the API will apply the changes when running the pipelineRun. However, this requires the pipelines to be generated from the SDK with uuid enabled to make sure no conflicts in our pipeline versioning.

**Additional information:**

Tomcli:
Resolves #496 

Since we don't have support for putting custom task in the task spec. The below diagram shows how we could offload some of the extra api calls from the users to KFP apiserver. Ideally, having taskSpec for custom task is a better solution but it will take a while before it can be used in a Tekton stable release. 

This PR handles the first step below where KFP apiserver creates the custom task CR x N for the users. Where N is the number of CRs and each CR needs a new API call.
Screen Shot 2021-03-24 at 6 41 07 PM

Although this feature allows users to deploy all the necessary custom resources with one KFP API call, KFP doesn't clean up these custom resources because the same custom resource could have multiple owners. This issue should be resolved in the upcoming taskSpec support for the custom task, so we don't want to create new dependencies on this temporary solution.

**Environment tested:**

* Python Version (use `python --version`): 3.7
* Tekton Version (use `tkn version`): 0.21.0
* Kubernetes Version (use `kubectl version`): 1.18

**Checklist:**


Tomcli:
A new use case for this is with recursion. Since recursion loop has to reference a loop resource, we cannot use taskSpec to replace this feature because taskSpec is not a referable resource that can call itself in the same pipeline.

sdk/python/tests/compiler/testdata/conditions_and_loops.yaml
```diff
@@ -33,6 +33,195 @@ metadata:
       [{"key": "artifacts/$PIPELINERUN/print-number/Output.tgz", "name": "print-number-Output",
       "path": "/tmp/outputs/Output/data"}], "produce-numbers": [{"key": "artifacts/$PIPELINERUN/produce-numbers/Output.tgz",
       "name": "produce-numbers-Output", "path": "/tmp/outputs/Output/data"}]}'
+    tekton.dev/resource_templates: "\n---\napiVersion: custom.tekton.dev/v1alpha1\n\
```
ckadner, line 36: Hi Tommy, do you think we could convert the resource template string to JSON/Dict here in order to preserve some readability? I don't presume many users will want/need to read this, but for debugging and understanding purposes JSON should be easier

Tomcli: @drewbutlerbb4 @fenglixa do you have any preference on JSON vs yaml when parsing strings in the UI? If not, I will convert all the custom task CRs into to list of JSON as @ckadner suggested. JSON also could help us save some character space as well.

Tomcli: This annotation has two purpose, 
1. API uses it to deploy the custom task CR before running the pipeline
2. UI able to use this information to display loop dag with more details.

fenglixa: @Tomcli, yes, I used the json format handled such situation before, refer to my another PR, FYI:
https://github.com/kubeflow/kfp-tekton/pull/489/files#diff-5aee3b3cd734f4dd317784331d99f1e76d0283631755c31cd67379220e883734R224-R226

Tomcli: done, the annotations now store the loop CRs as JSON string

Tomcli:
/hold
for reviews

ckadner (approved):
Thanks @Tomcli, I like it :-)

ckadner:
Good idea to put this PR on hold for others to review, so I can safely add my +1 while still giving others a chance for review 👍🏻 

/lgtm

Tomcli:
cc @Udiknedormin @rafalbigaj any comment/feedback on this PR? Right now we are using it for all the loop resources. However, after we support taskSpec for custom task, we will mainly use it on loop resources for recursion.

fenglixa:
After https://github.com/kubeflow/kfp-tekton/pull/561 merged. this PR may need to be changed, right?

Tomcli:
@fenglixa if #561 get merged first, then we will update this PR to deploy loop resources only for recursion.

fenglixa:
Great, Thanks @Tomcli

guides/kfp-admin-guide.md
```diff
@@ -93,6 +106,8 @@ KFP Caching will cache all the workloads that use the same task with the same in
 kubectl set env -n kubeflow deploy/ml-pipeline CACHE_ENABLED=false
 ```
 
+<<<<<<< HEAD
+=======
 ## Change the Tekton Terminate API Method
```
pugangxa, line 111: Resolve confict?

guides/kfp-admin-guide.md
```diff
@@ -110,4 +125,5 @@ kubectl rollout restart deploy/ml-pipeline -n kubeflow
 ```
 
 
+>>>>>>> upstream/master
```
pugangxa, line 128: And here.

guides/kfp-admin-guide.md
Tomcli: Thanks. Sorry I forgot to save my changes in my previous commit. I fixed it now.

guides/kfp-admin-guide.md
Tomcli: fixed it.

Tomcli:
@Udiknedormin We are planning to make this feature as default for recursion. Can you give a quick review to make sure this won't break your team's use case? Thanks.

Udiknedormin:
@Tomcli It looks like it works similar to what we already implemented, i.e. auto-creates resources (although it seems to be at a different time). It also doesn't remove the resources while I think our version does. If I understood correctly, it also requires moving resources to some annotation, which our version doesn't do nor need.

We would rather still use our version, but if it's possible to simply turn this new feature off, I think we should be fine with it.

Tomcli:
@Udiknedormin I added more details on how to disable this feature both from the API and SDK. The reason I didn't let kfp to remove the resources because there could be multiple pipelines that are referring to the same resource. If we think it is safe enough, I can add owner reference to those resources that bound to the latest pipelineRun that are using it.

Tomcli:
@pugangxa can you give a final review and lgtm this PR? Thanks.

Tomcli:
This feature should be enabled by default and can be opt-out with a flog

PR #539, opened 2021-04-14
Support the simple condition for any sequencer

### LINKED ISSUE #532: Any-Sequencer --- wait for some output to be true/false/some-value
The current implementation of AnySequencer, both in kfp-tekton as DSL extension and the controller itself, can only wait for multiple tasks to complete. It doesn't provide possibility for asynchronous "or" operation --- waiting for any condition to be fulfilled. That's needed to implement the Any-Sequencer's relationship with conditional nodes.

Considering how conditional links are evaluated as CELs and exposing `status` output, it would be enough for AnySequencer to modify its argument:

```python
class AnySequencer(ContainerOp):
  def __init__(
    self,
    any: Iterable[Union[dsl.ContainerOp, ConditionOperator]]
  ):
    ...
```

AnySequencer should wait for either one of its `dsl.ContainerOp` dependencies to finish execution, or for one of its `ConditionOperator` dependencies to become available (i.e. all of its tasks are completed) and to evaluate to `true`.

Example:

```python
long_calc_op = LongCalculationOp(some_inputs)
short_calc_op = ShortCalculationOp(some_inputs)
seq = AnySequencer(long_calc_op, short_calc_op.outputs["status"] == "success")
```

In the above example, `long_calc_op` takes long to calculate, but always returns valid result, while `short_calc_op` is some heuristic which is faster, but its results aren't always correct. Because of that, `seq` waits for either `long_calc_op` to finish, or for `short_calc_op` to finish with `"status"` output set to `"success"`.

pugangxa:
Resolves #532 

- DSL: When defining the any tasks to watch, user can specify a simple condition and it must meet for any sequencer to continue. 
- Image: refactored to support the condition


pugangxa:
/cc @jinchihe @Tomcli 

Just complete code, need more test, at least I can see issues for:
- For condition I split by "_" but the task names may contain "_" too
- Some assert since we only support pipeline parameter and primitive types

Tomcli:
@pugangxa FYI, in KFP, they call the code applied in the `.after()` as `extensions`. DSL is only referring changes in the kfp.dsl code.

sdk/python/tests/compiler/testdata/any_sequencer.yaml
```diff
@@ -123,12 +148,14 @@ spec:
...
-          - -namespace
+          - --namespace
           - $(params.pipelineRun-namespace)
-          - -prName
+          - --prName
           - $(params.pipelineRun-name)
-          - -taskList
+          - --taskList
           - task1,task2,task3
+          - -c
```
Tomcli, line 157: @pugangxa looks like it needs double dashes flag `--c`. Right now I'm seeing this error
```
flag provided but not defined: -c
Usage of any-taskrun:
  -namespace string
    	The namespace of the pipelinerun.
  -prName string
    	The name of the pipelinerun.
  -taskList string
    	The comma separated list of the tasks.
```

Tomcli: or maybe you haven't update the image?

pugangxa: Yes Tommy I haven't update the image, I should update it after code merged right?

pugangxa:
> @pugangxa FYI, in KFP, they call the code applied in the `.after()` as `extensions`. DSL is only referring changes in the kfp.dsl code.

Thanks for the information. I'm afraid I do not quite follow you, so this part keep the previous design and just add the condition clause. Is there anything change needed here?

sdk/python/kfp_tekton/tekton.py
```diff
@@ -31,34 +32,64 @@ class AnySequencer(ContainerOp):
...
         tasks_list = []
+        condition_list = []
         for cop in any:
-            cop_name = sanitize_k8s_name(cop.name)
-            tasks_list.append(cop_name)
+            if isinstance(cop, dsl.ContainerOp):
+                cop_name = sanitize_k8s_name(cop.name)
+                tasks_list.append(cop_name)
+            elif isinstance(cop, ConditionOperator):
+                condition_list.append(cop)
```
jinchihe, line 48: if not `dsl.ContainerOp` or `ConditionOperator`, raise a exception?

sdk/python/kfp_tekton/tekton.py
```diff
@@ -31,34 +32,64 @@ class AnySequencer(ContainerOp):
...
-                        task_list_str
-                    ],
+            arguments=arguments,
         )
 
 
-def after_any(any: List[dsl.ContainerOp], name: str = None):
+def processOperand(operand) -> str:
+    if isinstance(operand, dsl.PipelineParam):
+        return "results_" + operand.op_name + "_" + operand.name
```
jinchihe, line 72: can here to ensure the op_name is one of task in the pipelines? I meant if the task is not in the pipelines, not sure is there any problem?

tekton-catalog/any-sequencer/Makefile
```diff
@@ -24,10 +24,10 @@ init:
 	mkdir -p ${BIN_DIR}
 
 local: init
-	go build -o=${BIN_DIR}/any-taskrun ./cmd
+	go build -o=${BIN_DIR}/any-taskrun
 
 build-linux: init
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o=${BIN_DIR}/any-taskrun ./cmd
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o=${BIN_DIR}/any-taskrun
```
jinchihe, line 30: do we need to update the name `any-taskrun`? since we consider conditions.

tekton-catalog/any-sequencer/cmd/root.go
```diff
@@ -0,0 +1,246 @@
...
+		resultMatcher := regexp.MustCompile(`results_([^_]*)_([^_]+)`)
+		operand1Results := resultMatcher.FindAllStringSubmatch(operands[0], -1)
+		operand2Results := resultMatcher.FindAllStringSubmatch(operands[2], -1)
+		if len(operand1Results) == 0 && len(operand2Results) == 0 {
+			log.Printf("Must at least contain one result and at most two in one condition for a task.")
+			os.Exit(1)
+		}
+		if len(operand1Results) > 0 && len(operand2Results) > 0 {
+			if operand1Results[0][1] != operand2Results[0][1] {
+				log.Printf("The conditon can only contain results in one task, here's two.")
```
jinchihe, line 82: shall we ensure this while compiling in DSL? seems that's later in any seq pod.

tekton-catalog/any-sequencer/cmd/root.go
```diff
@@ -0,0 +1,246 @@
...
+							watcher.Stop()
+							os.Exit(0)
+						}
+					} else {
+						watcher.Stop()
+						os.Exit(0)
+					}
+				} else if taskrun.Status.GetCondition(apis.ConditionSucceeded).IsFalse() {
+					if !contains(failedTasks, taskLabel) {
+						failedTasks = append(failedTasks, taskLabel)
```
jinchihe, line 235: if so, the `taskrun` and `taskrun.condition` will be `all` relationship, but not any. I'm OK if this is design.

jinchihe (commented):
Overall looks great, thanks @pugangxa

sdk/python/kfp_tekton/tekton.py
pugangxa: Yes, need assert here.

tekton-catalog/any-sequencer/cmd/root.go
```diff
@@ -0,0 +1,246 @@
...
+		for event := range watcher.ResultChan() {
+			taskrun := event.Object.(*v1beta1.TaskRun)
+			taskLabel := taskrun.Labels["tekton.dev/pipelineTask"]
+			if contains(tasks, taskLabel) {
+				if taskrun.Status.GetCondition(apis.ConditionSucceeded).IsTrue() {
+					log.Printf("The TaskRun of %s succeeded.", taskLabel)
+					// check the conditions for the task
+					if conditions, ok := conditionMap[taskLabel]; ok {
+						if condition, ok := checkConditions(conditions, taskrun); !ok {
+							log.Printf("The condition %s for the task %s does not meet, still waiting", condition, taskLabel)
```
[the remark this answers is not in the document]
pugangxa: Sure

sdk/python/kfp_tekton/tekton.py
pugangxa: PipelineParam's op_name should be one of task in the pipelines, here we need assert the operand is either int or string or PipelineParam like I mentioned before.

tekton-catalog/any-sequencer/Makefile
pugangxa: I'm indifferent but think we should still keep it simple, while open to good names if have.

tekton-catalog/any-sequencer/cmd/root.go
pugangxa: In fact checked it in image, but I agree we need assert on DSL side.

tekton-catalog/any-sequencer/cmd/root.go
pugangxa: In fact the taskrun and the condition on the same taskrun is of ```all``` relationship. That's by design.

sdk/python/kfp_tekton/tekton.py
Udiknedormin: @pugangxa Why does the above use `"_"`? It seems to me that `"_"` is allowed for parameter names in Tekton, so wouldn't some forbidden character, like `"."` be better?

sdk/python/kfp_tekton/tekton.py
```diff
@@ -31,34 +32,64 @@ class AnySequencer(ContainerOp):
...
+    else:
+        return operand
+
+
+def processCondition(condition: ConditionOperator) -> str:
+    conditionStr = processOperand(condition.operand1)
+    conditionStr = conditionStr + " " + condition.operator + " "
+    conditionStr = conditionStr + processOperand(condition.operand2)
```
Udiknedormin, line 80: Looks a bit confusing. How about the following?

```python
op1 = processOperand(condition.operand1)
op2 = processOperand(condition.operand2)
conditionStr = f"{op1} {condition.operator} {op2}"
```

Tomcli:
> > @pugangxa FYI, in KFP, they call the code applied in the `.after()` as `extensions`. DSL is only referring changes in the kfp.dsl code.
> 
> Thanks for the information. I'm afraid I do not quite follow you, so this part keep the previous design and just add the condition clause. Is there anything change needed here?

The design is okay. Just that the KFP team is sensitive when we are calling something as "DSL" changes in the design, so we try to avoid using the term DSL if possible. Previously, I moved the any sequencer to the tekton.py file because the any sequencer is considered as `extensions` in KFP.

sdk/python/kfp_tekton/tekton.py
```diff
@@ -31,34 +32,64 @@ class AnySequencer(ContainerOp):
...
+                    "$(context.pipelineRun.name)",
+                    "--taskList",
+                    task_list_str
+                ]
+        conditonArgs = processConditionArgs(condition_list)
+        arguments.extend(conditonArgs)
 
         super().__init__(
             name=name,
             image="dspipelines/any-sequencer:latest",
```
Tomcli, line 64: Can we make this image name configurable? This way is easier to run some testing in our internal CI pipelines.

tekton-catalog/any-sequencer/cmd/root.go
```diff
@@ -0,0 +1,247 @@
...
+							failedTasks = append(failedTasks, taskLabel)
+						} else {
+							watcher.Stop()
+							os.Exit(0)
+						}
+					} else {
+						watcher.Stop()
+						os.Exit(0)
+					}
+				} else if taskrun.Status.GetCondition(apis.ConditionSucceeded).IsFalse() {
```
Udiknedormin, line 233: That's a very deeply-indented block of code, it's a bit difficult to read.

Go idiom is for `if`s to handle fast-exit behaviours, while handling the other ones outside of the `if`. Considering how you call `os.Exit(0)` here, it seems possible to do it the idiomatic way:

```go
taskrunStatus = taskrun.Status.GetCondition(apis.ConditionSucceeded)

if taskrunStatus.IsTrue() {
  conditions, ok := conditionMap[taskLabel]
  if !ok { // no conditions to be passed --> any-sequencer success
    watcher.Stop()
    os.Exit(0)
  }

  condition, ok := checkConditions(conditions, taskrun)
  if ok { // condition passed -->  any-sequencer success
    watcher.Stop()
    os.Exit(0)
  }

  log.Printf("The condition %s for the task %s does not meet.", condition, taskLabel)
}

if !contains(failedTasks, taskLabel) {
  failedTasks = append(failedTasks, taskLabel)
}

if len(failedTasks) >= len(tasks) {
  log.Printf("All specified TaskRun(s) failed.")
  watcher.Stop()
  os.Exit(1)
}
```

sdk/python/kfp_tekton/tekton.py
pugangxa: Yes this is one of the places need fix I mentioned in comment. It's a little tricky because some characters that's allowed in Tekton but not allowed for goevaluate, and some in goevaluate not allowed in Tekton. I don't want to translate between goevaluate and Tekton so maybe use the format like ```taskName_output_outputName_end``` which is easy to parse and allowed by both.

sdk/python/kfp_tekton/tekton.py
pugangxa: Sure, will do.

tekton-catalog/any-sequencer/cmd/root.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		cr := conditionResult{
+			condition: condition,
+			results:   resultName,
+		}
+		if len(conditionMap[taskName]) != 0 {
+			conditionMap[taskName] = append(conditionMap[taskName], cr)
+		} else {
+			conditionMap[taskName] = []conditionResult{cr}
+		}
```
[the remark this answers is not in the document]
pugangxa: OK

tekton-catalog/any-sequencer/cmd/root.go
```diff
@@ -0,0 +1,247 @@
...
+		Run: watch,
+	}
+
+	rootCmd.Flags().StringVar(&namespace, "namespace", "", "The namespace of the pipelinerun.")
+	rootCmd.MarkFlagRequired("namespace")
+	rootCmd.Flags().StringVar(&prName, "prName", "", "The name of the pipelinerun.")
+	rootCmd.MarkFlagRequired("prName")
+	rootCmd.Flags().StringVar(&taskList, "taskList", "", "The comma separated list of the tasks.")
+	rootCmd.MarkFlagRequired("taskList")
+	rootCmd.Flags().StringSliceVarP(&conditions, "condition", "c", []string{}, "The condition to watch")
```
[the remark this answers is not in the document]
pugangxa: OK, will modify the message.

tekton-catalog/any-sequencer/cmd/root.go
pugangxa: Thanks for your suggestion, I just modified it based on original code, I agree it's more concise this way, thanks!

sdk/python/kfp_tekton/tekton.py
pugangxa: Just checked that "_" is not allowed  for parameter names in Tekton, in kfp-tekton all "_" will be replaced with "-", so we can still use it but must process '-' in goevaluate since it's not allowed in goevaluate

tekton-catalog/any-sequencer/cmd/root.go
pugangxa: Tested and we cannot do this, the reason is for the task if the status is running or unknown, it will also be appended to failed task. We should only append to failed task list if the task status is false or the condition failed.

pugangxa:
@fenglixa @Tomcli All the comments addressed, please help approve and QA will do more testing, thanks. After code merged I will update image.

fenglixa:
@jinchihe , is this code diff OK to you?

jinchihe:
/lgtm
Great work! Thanks @pugangxa

PR #544, opened 2021-04-15
(ui) Update UI to render custom tasks

### LINKED ISSUE #535: Update UI to render user defined Tekton custom task
/kind feature

**Description:**
We introduced new features that allow users to define Tekton custom task in KFP. However, the UI needs to be updated in order to render any user defined custom task.
e.g. we need the UI to render this condition example
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/testdata/condition_custom_task.yaml

**Additional information:**

drewbutlerbb4:
Resolves #535 

Add parsing for custom tasks in Pipeline Details and Run Details pages. This includes parsing the and adding custom tasks to the static graph structure and the dynamic graph structure, parsing the input params, and  parsing the task results for custom tasks.

**Environment tested:**

**Checklist:**


frontend/src/lib/WorkflowParser.ts
```diff
@@ -75,6 +75,13 @@ export default class WorkflowParser {
         statusMap.set(status[taskRunId]['pipelineTaskName'], status[taskRunId]);
     }
 
+    // Add custom tasks to the status map
+    for (const taskRunId of Object.getOwnPropertyNames(workflow['status']['runs'])) {
```
Tomcli, line 79: Object.getOwnPropertyNames() may return nil instead of empty array. E.g. you will get an error when running the cache example. https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/testdata/cache.py

Tomcli: need to fall back to an empty array if Object.getOwnPropertyNames() == nil

frontend/src/lib/WorkflowParser.ts
```diff
@@ -309,6 +317,16 @@ export default class WorkflowParser {
       }
     }
 
+    // Loop for custom tasks
+    for (const runId of Object.getOwnPropertyNames(workflow.status.runs)) {
```
Tomcli, line 321: Object.getOwnPropertyNames can be nil over here as well

drewbutlerbb4:
Thanks for the catch @Tomcli 

With the addition of custom tasks is there still a hard requirement for "normal" taskRuns to be present? Or could a pipeline be made entirely of custom tasks?

Tomcli:
@drewbutlerbb4 custom tasks don't have `taskRuns`, they are using the `Runs` CRD as the equivalent to track their runtime status. And yes, a pipeline could potentially made with all custom tasks.

Tomcli:
So we don't need a hard requirement to have a taskrun for each node/pipeline.

drewbutlerbb4:
Thanks @Tomcli 
I have added a check for Object.getOwnPropertyNames() == nil for status.taskRuns to account for that scenario

Tomcli:
Thanks @drewbutlerbb4. This looks good. However, there's some edge case errors in the dynamic graph for recursion. I will open another issue to summarize the recursion edge case since that's outside of the scope for #544

PR #550, opened 2021-04-20
Adjust the any sequencer example to make it more reasonable

pugangxa:
I uploaded image and did E2E test. Found the example is not reasonable, because the flip-coin to show the condition will never run because task1 just succeed quickly.

Here update the example then it make more sense, thanks.


Tomcli:
thanks @pugangxa, can you also summarize the condition usage and its limitation (like the succeeded too quickly in this example) at the feature.md? thanks 
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/FEATURES.md#any-sequencer

pugangxa:
> thanks @pugangxa, can you also summarize the condition usage and its limitation (like the succeeded too quickly in this example) at the feature.md? thanks
> https://github.com/kubeflow/kfp-tekton/blob/master/sdk/FEATURES.md#any-sequencer

Sure, will do.

Issue #553, opened 2021-04-20
Document the any sequencer condition usage and its limitation at feature.md


Tomcli:
@pugangxa, can you also summarize the any sequencer condition usage and its limitation (like the succeeded too quickly in this example) at the feature.md? thanks 
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/FEATURES.md#any-sequencer

_Originally posted by @Tomcli in https://github.com/kubeflow/kfp-tekton/issues/550#issuecomment-823419287_

Tomcli:
/close
by #557

Issue #552, opened 2021-04-20
[Any-Sequencer][Condition]any-test cannot exit when the condition does not meet.


huixa:
**Test the any-sequencer condition #539 , some output messages which it's better to enhance:**

**Env:**
build the env for any-sequencer from kfp-tekton/tekton-catalog/any-sequencer

**Issue:**
When the condition does not meet, the any-test will watch forever, don't exit.

**Configure:**
    - name: any-test
      taskSpec:
        steps:
        - name: main
          args:
          - --namespace
          - $(params.pipelineRun-namespace)
          - --prName
          - $(params.pipelineRun-name)
          - --taskList
          - 
          - -c
          - results_flipcoin_output1 == 'heads'
          
**Result:**
[root@symtest21 test]# kubectl -n kubeflow logs any-sequencer-any-test-wh6ds-pod-6s4hx
time="2021-04-20T06:03:13Z" level=info msg="Starting to watch taskrun for '' and condition in kubeflow/any-sequencer."
time="2021-04-20T06:03:30Z" level=info msg="The TaskRun of flipcoin succeeded."
time="2021-04-20T06:03:30Z" level=info msg="The result output1 does not exist in taskrun flipcoin."
time="2021-04-20T06:03:30Z" level=info msg="The condition results_flipcoin_output1 == 'heads' for the task flipcoin does not meet."
[root@symtest21 test]# 

[root@symtest21 test]# kubectl -n kubeflow get po
NAME                                               READY   STATUS      RESTARTS   AGE
any-sequencer-any-test-wh6ds-pod-6s4hx             1/1     Running     0          8m6s
any-sequencer-flipcoin-9699p-pod-g29r4             0/1     Completed   0          7m48s
any-sequencer-task1-srsv6-pod-f5842                0/1     Completed   0          8m6s
any-sequencer-task2-x24s5-pod-m44zt                0/1     Completed   0          8m6s
any-sequencer-task3-fgsf7-pod-lb9jb                0/1     Completed   0          8m6s

**Expect Result:**
When does not meet the condition or does not find the condition in taskrun, should exit the any-test.

huixa:
@pugangxa , please help handle this one. Thanks.

pugangxa:
/assign
OK will take a look, thanks!

Udiknedormin:
[...]

PR #557, opened 2021-04-21
Fix multiple any-sequencer issues

### LINKED ISSUE #551: [Any-Sequencer][Condition]Enhance the output messages.
**Test the any-sequencer condition #539 , some output messages which it's better to enhance:**

**Env:**
build the env for any-sequencer from kfp-tekton/tekton-catalog/any-sequencer

===================
**Update 1:**
**When trigger the any by one condition, the output message as below:**
_[root@symtest21 test]# kubectl -n kubeflow logs any-sequencer-any-test-lnbrq-pod-w4698
time="2021-04-20T05:40:28Z" level=info msg="Starting to watch taskrun for 'task1,task3' and condition in kubeflow/any-sequencer."
time="2021-04-20T05:40:47Z" level=info msg="The TaskRun of flipcoin succeeded."
[root@symtest21 test]#_

**Suggestion:**
1)Whether could add the condition information in the first line's log:
"Starting to watch taskrun for 'task1,task3' and condition in kubeflow/any-sequencer."

2)Whether could add the related messages when the meet the condition.

pugangxa:
1: Add log when condition check suceeded.
2: Fix the issue when task list is not provided.
3: Fix the issue when there's multiple results provided
4: Update the document

Resolves #551 #552 #553

### DIFF
--- a/sdk/FEATURES.md
+++ b/sdk/FEATURES.md
@@ -142,7 +142,7 @@ To see how the Python SDK provides this feature, refer to the examples below:
 
 ### Any Sequencer
 
-When any one of the task dependencies completes successfully, the dependent task will be started. Order of execution of the dependencies doesn’t matter, and the pipeline doesn't wait for all the task dependencies to complete before moving to the next step. Please follow the details of the implementation in the [design doc](https://docs.google.com/document/d/1oXOdiItI4GbEe_qzyBmMAqfLBjfYX1nM94WHY3EPa94/edit#heading=h.dt8bhna4spym).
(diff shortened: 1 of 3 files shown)


pugangxa:
@jinchihe @Tomcli Please help task a look, thanks.

sdk/FEATURES.md
```diff
@@ -142,7 +142,7 @@ To see how the Python SDK provides this feature, refer to the examples below:
...
+When any one of the task dependencies completes successfully and the conditions meet, the dependent task will be started. Order of execution of the dependencies doesn’t matter, and the pipeline doesn't wait for all the task dependencies to complete before moving to the next step. Condition can be applied to enforce the task dependencies completes as expected. The conditon expression should be the same format as is in Kubeflow ConditionOperator, and the result of containerOps can be used in expression. Notice the expression should only contain result from only one task because the purpose here is to check the simple condition for the task's output when a task complete. And also the operand in the expression should be int or string, other python types will be transferred to string automatically. Please follow the details of the implementation in the [design doc](https://docs.google.com/document/d/1oXOdiItI4GbEe_qzyBmMAqfLBjfYX1nM94WHY3EPa94/edit#heading=h.dt8bhna4spym).
```
Tomcli, line 145:
```suggestion
When any one of the task dependencies completes successfully and the conditions meet, the dependent task will be started. Order of execution of the dependencies doesn’t matter, and the pipeline doesn't wait for all the task dependencies to complete before moving to the next step. Condition can be applied to enforce the task dependencies completes as expected. The condition expression should be the same format as is in Kubeflow ConditionOperator, and the result of containerOps can be used in expression. Notice the expression should only contain results from only one task because the purpose here is to check the simple condition for the task's output when a task complete. And also the operand in the expression should be int or string, other python types will be transferred to string automatically. Please follow the details of the implementation in the [design doc](https://docs.google.com/document/d/1oXOdiItI4GbEe_qzyBmMAqfLBjfYX1nM94WHY3EPa94/edit#heading=h.dt8bhna4spym).
```

pugangxa: Thanks Tommy, updated.

PR #560, opened 2021-04-21
Update katib sample

yhwang:
Update the notebook to use newer packages and image.
Also add instructions for multi-user environment.

### DIFF
--- a/samples/katib/early-stopping.ipynb
+++ b/samples/katib/early-stopping.ipynb
@@ -29,212 +29,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Defaulting to user installation because normal site-packages is not writeable\n",
-      "Collecting pip\n",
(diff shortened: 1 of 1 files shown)


samples/katib/early-stopping.ipynb
```diff
@@ -29,212 +29,16 @@
...
    "source": [
     "# Update the PIP version.\n",
     "!python -m pip install --upgrade pip\n",
-    "!pip install kfp==1.1.1\n",
-    "!pip install kubeflow-katib==0.10.1\n",
-    "!pip install kfp-tekton==0.4.0"
+    "!pip install kfp==1.4.0\n",
+    "!pip install kubeflow-katib==0.11.0\n",
+    "!pip install kfp-tekton==0.7.0\n",
+    "!pip install kubernetes==12.0.0"
```
Tomcli, line 41: Maybe add a command to restart the kernel after this? Since kubeflow-katib may override the kubernetes client version.
```
from IPython.core.display import HTML
HTML("<script>Jupyter.notebook.kernel.restart()</script>")
```

yhwang: it already has a `restart` section below :-)

Tomcli: oh sorry didn't see that, thanks

PR #589, opened 2021-05-17
Add Parallelism support on backend.

### LINKED ISSUE #542: Support parallelism for pipeline loop
/kind feature
Introduce the parallism parameter to control the concurrency

**Description:**
Before we have already implemented loop, but all the pipeline can only run in sequential. We need support parallism parameter so that multiple pipelines can be started at once with one loop arg for each one.

pugangxa:
- Parallelism is defaulted to 1
- if Parallelism is bigger then total iteration numbers, then start all
- normally start the same number of pipelineruns as Parallelism value

Resolves #542

### DIFF
--- a/tekton-catalog/pipeline-loops/pkg/apis/pipelineloop/v1alpha1/pipelineloop_defaults.go
+++ b/tekton-catalog/pipeline-loops/pkg/apis/pipelineloop/v1alpha1/pipelineloop_defaults.go
@@ -18,6 +18,8 @@ package v1alpha1
 
 import (
 	"context"
+	"os"
+	"strconv"
 
 	"knative.dev/pkg/apis"
 )
@@ -31,4 +33,19 @@ func (tl *PipelineLoop) SetDefaults(ctx context.Context) {
 
 // SetDefaults set any defaults for the PipelineLoop spec
 func (tls *PipelineLoopSpec) SetDefaults(ctx context.Context) {
+	if tls.Parallelism == 0 {
+		parallelism := os.Getenv("LOOP_PARALLELISM")
+		if parallelism == "" {
+			tls.Parallelism = 1
+			return
+		}
+		i, err := strconv.Atoi(parallelism)
+		if err != nil {
+			// fall back to default 1
+			tls.Parallelism = 1
+			return
+		} else {
+			tls.Parallelism = i
+		}
+	}
 }
--- a/tekton-catalog/pipeline-loops/README.md
+++ b/tekton-catalog/pipeline-loops/README.md
@@ -31,6 +31,8 @@
 
 - Install pipelineloop controller `kubectl apply -f config/`
 
+# Parallelism
+`Parallelism` is to define the number of pipelineruns can be created at the same time. It must be eqaul to or bigger than 1. If it's not set then the default value will be 1. If it's bigger than the total iterations in the loop then the number will be total iterations.
 
 # Verification
 - check controller and the webhook. `kubectl get po -n tekton-pipelines`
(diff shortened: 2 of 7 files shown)


pugangxa:
/hold
Just finished code, need test, please help take a look firstly. @fenglixa @Tomcli

fenglixa:
Thanks @pugangxa , 
Would you please add some examples here?
https://github.com/kubeflow/kfp-tekton/tree/master/tekton-catalog/pipeline-loops/examples

pugangxa:
> Thanks @pugangxa ,
> Would you please add some examples here?
> https://github.com/kubeflow/kfp-tekton/tree/master/tekton-catalog/pipeline-loops/examples

Sure, I will add test and examples, and also I am doing some basic testing. Thanks.

tekton-catalog/pipeline-loops/pkg/apis/pipelineloop/v1alpha1/pipelineloop_defaults.go
```diff
@@ -31,4 +31,7 @@ func (tl *PipelineLoop) SetDefaults(ctx context.Context) {
 
 // SetDefaults set any defaults for the PipelineLoop spec
 func (tls *PipelineLoopSpec) SetDefaults(ctx context.Context) {
+	if tls.Parallelism == 0 {
+		tls.Parallelism = 1
```
Tomcli, line 35: Can we set the default Parallelism number as a environment variable? This way the admin can change it based on the cluster size.

pugangxa: Do you mean to make it an environment variable for webhook? Though user can set it purposely but I agree it's always good to have a suitable default value, I can do that.

pugangxa:
/unhold
Basic cases tested. @Tomcli @fenglixa

pugangxa:
Will make another PR for DSL changes to avoid messed up.

tekton-catalog/pipeline-loops/pkg/apis/pipelineloop/v1alpha1/pipelineloop_defaults.go
```diff
@@ -31,4 +33,19 @@ func (tl *PipelineLoop) SetDefaults(ctx context.Context) {
 
 // SetDefaults set any defaults for the PipelineLoop spec
 func (tls *PipelineLoopSpec) SetDefaults(ctx context.Context) {
+	if tls.Parallelism == 0 {
```
Tomcli, line 36: When I explicitly set Parallelism: 0 in my pipelineloop, it will still fall back to the default because this condition will pass for both null and 0. Shouldn't if a user specify `Parallelism: 0`, then the controller won't limit the number of parallel pipelineruns?

Tomcli (commented):
The parallelism feature looks good to me, just one question on setting the default parallelism value.

tekton-catalog/pipeline-loops/pkg/apis/pipelineloop/v1alpha1/pipelineloop_defaults.go
pugangxa: Thanks Tommy，in fact I am intertwined with this default value too. The original design is to use 0 to represent unlimited. But you know in golang the default value for int is 0 which means if user did not specify it then it will be unlimited. Originally it's 1 without this feature so this maybe a surprise to user so I think we should fall back to default 1? BTW, in golang it's still 0 for unlimited and I will change it. What's your opinion?

pugangxa: Addressed and readme updated based on our discussion, thanks.

fenglixa:
Thanks @pugangxa 
/lgtm
/approve

PR #592, opened 2021-05-19
Support parallelism feature on DSL side

### LINKED ISSUE #542: Support parallelism for pipeline loop

pugangxa:
Resolves #542 

Add parallelism to yaml spec.


pugangxa:
@fenglixa @Tomcli 
Please help take a look, thanks.

sdk/python/tests/compiler/testdata/loop_static_with_parallelism.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - args:
+          - echo $(inputs.params.my_pipe_param)
+          command:
+          - sh
+          - -c
+          image: library/bash:4.4.23
+          name: main
+      timeout: 0s
+    - name: para-loop-pipeline-for-loop-2
+      parallelism: 5
```
Tomcli, line 63: I don't think the `parallelism` is valid over here. Although the Tekton controller will ignore it for now, it might cause problems in the future if Tekton check this field.

sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -140,6 +140,17 @@ def test_sidecar_workflow(self):
...
+  def test_loop_parallelism_workflow(self):
+    """
+    Test compiling a loop with parallelism defined workflow.
+    """
+    from .testdata.loop_static_with_parallelism import pipeline
+    self._test_pipeline_workflow(
+      pipeline,
+      'loop_static_with_parallelism.yaml',
+      normalize_compiler_output_function=lambda f: re.sub(
+          "(loop-item-param-[a-z 0-9]*|for-loop-for-loop-[a-z 0-9]*)", "with-item-name", f))
```
Tomcli, line 152: `with-item-name` is not a valid parameter for this pipeline, is there a reason you need to substitute over here?

pugangxa: Oops, I just copied other test cases and forgot to delete this, thanks.

sdk/python/tests/compiler/testdata/loop_static_with_parallelism.yaml
pugangxa: OK, I will remove this from the pipelinerun yaml.

pugangxa:
All the comments addressed. @Tomcli @fenglixa

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -393,6 +393,8 @@ def _group_to_dag_template(self, group, inputs, outputs, dependencies, pipeline_
               self.loops_pipeline[group_name]['spec']['params'].append({
                 'name': param[0], 'value': '$(params.%s)' % param[0]
               })
+      if sub_group.parallelism is not None:
```
fenglixa, line 396: Do you need to check wether is it int type for the value parallelism?

pugangxa: Thanks Feng，this parameter is a shared parameter with Argo, it's already validated in https://github.com/kubeflow/pipelines/blob/cc83e1089b573256e781ed2e4ac90f604129e769/sdk/python/kfp/dsl/_ops_group.py#L122-L126

fenglixa:
Thanks @pugangxa 
/lgtm
/approve

PR #583, opened 2021-05-07
Enhance Any Sequencer to support watching run

jinchihe:
**Which issue is resolved by this Pull Request:** 
Resolves #

**Description of your changes:**
Currently the Any Sequencer support watching taskrun and conditions, do not support watching run, the PR is going to support that.

**Environment tested:**

* Python Version (use `python --version`):
* Tekton Version (use `tkn version`):
* Kubernetes Version (use `kubectl version`):
* OS (e.g. from `/etc/os-release`):

**Checklist:**
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention). 

- [ ] Do you want this pull request (PR) cherry-picked into the current release branch?

    [Learn more about cherry-picking updates into the release branch](https://github.com/kubeflow/pipelines/blob/master/RELEASE.md#cherry-picking-pull-requests-to-release-branch).

PR #596, opened 2021-05-21
Permit pipeline-runner to operate on runs

pugangxa:
After PR #583, any-sequencer can watch runs but during test using pipeline-runner it complains lack of permission.This PR solve the issue. 
BTW, not sure whether we should add it for kfp-tekton.yaml under install directory. @Tomcli


Tomcli:
it's ok to add this to pipeline-runner since Argo also gives the same permission, although i will recommend to run with multi-user mode for a more secure environment. 
/lgtm
/approve

PR #665, opened 2021-07-10
Fix the yaml dump issue of boolean value

### LINKED ISSUE #664: Single character parameters need to have double quote to avoid yaml keyword conflicts
I think this is a bug, we should have double quote on single character parameter for n and y to avoid this issue. For now I will lgtm and open another issue for it

_Originally posted by @Tomcli in https://github.com/kubeflow/kfp-tekton/pull/660#discussion_r666403707_

pugangxa:
Resolves #664 

In test to dump yaml should use the same method as is in compile which is customized to process the boolean values.

### DIFF
--- a/sdk/python/tests/compiler/compiler_tests.py
+++ b/sdk/python/tests/compiler/compiler_tests.py
@@ -25,6 +25,7 @@ from os import environ as env
 
 import pytest
 import yaml
+from kfp_tekton.compiler.yaml_utils import dump_yaml
 from kfp_tekton import compiler
 from kfp_tekton.compiler.pipeline_utils import TektonPipelineConf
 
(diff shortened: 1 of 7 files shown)


Tomcli:
cc @Udiknedormin

Tomcli:
@pugangxa can you also update this file with parameter name `n`? thanks
https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/testdata/conditions_with_global_params.py

pugangxa:
> @pugangxa can you also update this file with parameter name `n`? thanks
> https://github.com/kubeflow/kfp-tekton/blob/master/sdk/python/tests/compiler/testdata/conditions_with_global_params.py

Sure, updated.

Issue #670, opened 2021-07-21
Migrate components away from containerOp


Tomcli:
/kind feature

**Description:**
We need to update all our helper functions and examples to use component.yaml instead of containerOp because containerOp is going to be removed in the v2 release.

- [ ] [Hyperparameter tuning using Katib](/samples/katib)
- [ ] [Training and Serving Models with Watson Machine Learning](/samples/watson-train-serve#training-and-serving-models-with-watson-machine-learning)
- [x] [Lightweight python components example](/samples/lightweight-component)
- [x] [Pipeline with Nested loops](/samples/nested-loops)
- [x] [Using Tekton Custom Task on KFP](/samples/tekton-custom-task)
- [x] [The flip-coin pipeline using custom task](/samples/flip-coin-custom-task)

**Additional information:**
[Miscellaneous information that will assist in solving the issue.]

Tomcli:
@yhwang can you take care of 
- [ ] https://github.com/kubeflow/kfp-tekton/tree/master/samples/nested-pipeline
- [ ] https://github.com/kubeflow/kfp-tekton/tree/master/samples/flip-coin
- [ ] https://github.com/kubeflow/kfp-tekton/tree/master/samples/trusted-ai

some of them should be partially using component.yaml already

yhwang:
try to note down the issues while updating testcases and samples:
- not sure how to use component.yaml with `dsl.InputArgumentPath`
  - sdk/python/tests/compiler/testdata/input_artifact_raw_value.py
- NotImplementedError: dsl.graph_component is not yet supported in KFP v2 compiler
   - sdk/python/tests/compiler/testdata/long_recursive_group_name.py
   - sdk/python/tests/compiler/testdata/loop_in_recursion.py
   - sdk/python/tests/compiler/testdata/nested_recur_custom_task.py
   - sdk/python/tests/compiler/testdata/nested_recur_runafter.py

Udiknedormin:
@Tomcli @yhwang 
What's the status of this issue?

Tomcli:
@Udiknedormin 
All the current examples in KFP-Tekton are migrated to component.yaml. The two examples that didn't check are the also in component.yaml but their service API was broken when we were testing it. Although containerOp is deprecated, it will continue to work in V1 KFP and KFP-Tekton.

PR #685, opened 2021-08-01
remove ContainerOp from test cases - part 1

yhwang:
update the testcase accordingly.

related to #670

remove ContainerOp and use Component instead

### DIFF
--- a/sdk/python/tests/compiler/testdata/affinity.py
+++ b/sdk/python/tests/compiler/testdata/affinity.py
@@ -13,16 +13,22 @@
 # limitations under the License.
 
 from kubernetes.client import V1Affinity, V1NodeSelector, V1NodeSelectorRequirement, V1NodeSelectorTerm, V1NodeAffinity
-from kfp import dsl
+from kfp import dsl, components
 
+ECHO_OP_STR = """
+name: echo
+description: echo component
+implementation:
+  container:
+    image: busybox
+    command:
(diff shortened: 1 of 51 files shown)


Tomcli:
thanks @yhwang, can you rebase this PR to the master so we can trigger the latest tests? Also, anything that is not working with the v2 compiler?

yhwang:
@Tomcli thanks for reminding me the v2 compiler. I double checked all the python files I changed in this PR. Only `exit_handler.py` fails when using v2 compiler. the error  is:
```
AssertionError: component_input_parameter: pipelineparam--pipelineparam--url not found. All inputs: 
```

the reason should be:
```
  A task input may reference an output outside its immediate DAG.
  For instance::

    random_num = random_num_op(...)
    with dsl.Condition(random_num.output > 5):
      print_op('%s > 5' % random_num.output)

  In this example, `dsl.Condition` forms a sub-DAG with one task from `print_op`
  inside the sub-DAG. The task of `print_op` references output from `random_num`
  task, which is outside the sub-DAG. When compiling to IR, such cross DAG
  reference is disallowed. So we need to "punch a hole" in the sub-DAG to make
  the input available in the sub-DAG component inputs if it's not already there,
  Next, we can call this method to fix the tasks inside the sub-DAG to make them
  reference the component inputs instead of directly referencing the original
  producer task.
```

PR #688, opened 2021-08-03
fix(sdk): add test for taskRef-vs-taskSpec params

### LINKED ISSUE #687: [SDK] No params passing for --taskSpec custom-tasks
/kind bug

Reported to me by Kamil Huras from our team.

He passed params to a custom-task specified with --taskRef and --taskSpec option.

At least in some cases, for --taskRef the params have been generated:

```python
dsl.ContainerOp(
        name=custom_task_name,
        image=custom_task_image,
        command=[custom_task_command],
        arguments=[
            "--apiVersion", custom_task_api_version,
            "--kind", custom_task_kind,
            "--name", custom_task_resource_name,
            *custom_task_args,
            "--taskRef", custom_task_resource
        ],
        file_outputs={
            f"{result}": f"/{result}"
            for result in custom_task_results
        }
    )
```
results in:

```yaml
kind: PipelineRun
metadata:
  name: (...)
  annotations: (...)
spec:
  params:
  - name: foo-bar
    value: 'buzz'
  pipelineSpec:
    params:
    - name: foo-bar
      default: 'buzz'
    tasks:
    - name: upd
      params:
      - name: foo-bar
        value: '$(params.foo-bar)'
      taskRef:
        apiVersion: (...)
        kind: (...)
        name: (...)
  timeout: 0s
```

...but for --taskSpec they are not:

```python
dsl.ContainerOp(
        name=cel_name,
        image=CEL_TASK_IMAGE_NAME,
        command=["cel"],
        arguments=[
            "--apiVersion", CEL_API,
            "--kind", CEL_KIND,
            "--name", cel_resource_name,
            *cel_args,
            "--taskSpec", resource
        ],
        file_outputs={
            f"{result}": f"/{result}"
            for result in results
        }
    )
```
results in:
```yaml
kind: PipelineRun
metadata:
  name: (...)
  annotations: (...)
spec:
  params:
  - name: foo-bar
    value: 'buzz'
  pipelineSpec:
    params:
    - name: foo-bar
      default: 'buzz'
    tasks:
    - name: (...)
      taskSpec:
        apiVersion: (...)
        kind: (...)
        spec:
          params:
          - name: params-foo-bar
            alias: params.foo_bar
            type: string
          (...)
        metadata:
          labels:
            pipelines.kubeflow.org/pipelinename: ''
            pipelines.kubeflow.org/generation: ''
            pipelines.kubeflow.org/cache_enabled: "true"
          annotations:
            valid_container: "false"
            tekton.dev/template: ''
  timeout: 0s
```
Note how there are no `params` in the task, only in its `taskSpec.spec` (i.e. they are declared but not instantiated).

--taskRef and --taskSpec only differing in where the resource is generated (in separate file or annotation vs inline), not in whenever either one can be passed parameters.

**Additional information:**
Bug in the `_data_passing_rewriter.py`:
```python
if 'condition-' not in task['name'] and not task.get('taskRef', '')
```
while it should be:
```python
if 'condition-' not in task['name'] and not task.get('taskRef', '') and not task.get('taskSpec', ''):
```

**Environment:**

* Python Version (use `python --version`): 3.9.0
* SDK Version: sha 186a9e301f2fa5e2bd11e4f86f6cafb97c123db7
* Tekton Version (use `tkn version`): irrelevant
* Kubernetes Version (use `kubectl version`): irrelevant
* OS (e.g. from `/etc/os-release`): irrelevant

Udiknedormin:
Resolves #687

Adds "taskSpec" field to the ones that shouldn't stop a custom-task from accepting params.

**Environment tested:**

* Python Version (use `python --version`): 3.9.0
* Tekton Version (use `tkn version`): irrelevant
* Kubernetes Version (use `kubectl version`): irrelevant
* OS (e.g. from `/etc/os-release`): irrelevant

**Checklist:**

- [ x Do you want this pull request (PR) cherry-picked into the current release branch?

I think it should be cherry-pickable.

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -341,7 +341,7 @@ def fix_big_data_passing(workflow: dict) -> dict:
     # Remove pipeline task parameters unless they're used downstream
     for task in pipeline_tasks:
         # Don't process condition and custom task parameters
-        if 'condition-' not in task['name'] and not task.get('taskRef', ''):
+        if 'condition-' not in task['name'] and not task.get('taskRef', '') and not task.get('taskSpec', ''):
             task['params'] = [
                 parameter_argument
                 for parameter_argument in task.get('params', [])
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -341,7 +341,9 @@ def fix_big_data_passing(workflow: dict) -> dict:
     # Remove pipeline task parameters unless they're used downstream
     for task in pipeline_tasks:
         # Don't process condition and custom task parameters
-        if 'condition-' not in task['name'] and not task.get('taskRef', '') and not task.get('taskSpec', ''):
+        is_condition = 'condition-' in task['name']
+        is_custom_task = task.get('taskRef', '') or task.get('taskSpec', '')
+        if not is_condition and not is_custom_task:
             task['params'] = [
                 parameter_argument
                 for parameter_argument in task.get('params', [])
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -342,7 +342,7 @@ def fix_big_data_passing(workflow: dict) -> dict:
     for task in pipeline_tasks:
         # Don't process condition and custom task parameters
         is_condition = 'condition-' in task['name']
-        is_custom_task = task.get('taskRef', '') or task.get('taskSpec', '')
+        is_custom_task = task.get('taskRef') or task.get('taskSpec', {}).get('apiVersion')
         if not is_condition and not is_custom_task:
             task['params'] = [
                 parameter_argument
--- a/sdk/python/tests/compiler/testdata/custom_task_params.py
+++ b/sdk/python/tests/compiler/testdata/custom_task_params.py
@@ -1,3 +1,17 @@
+# Copyright 2020 kubeflow.org
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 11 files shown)


sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -341,7 +341,7 @@ def mark_upstream_ios_of_output(template_output, marked_inputs,
     # Remove pipeline task parameters unless they're used downstream
     for task in pipeline_tasks:
         # Don't process condition and custom task parameters
-        if 'condition-' not in task['name'] and not task.get('taskRef', ''):
+        if 'condition-' not in task['name'] and not task.get('taskRef', '') and not task.get('taskSpec', ''):
```
Tomcli, line 344: `not task.get('taskSpec', '')` will also include regular tasks in this case, so what we need to do here is not to process a task that has taskSpec with custom task. Unfortunately in Tekton yaml, there's no concept of API type like in goLang, so we probably need to check some special annotations (e.g. `valid_container`) to determine whether the taskSpec is for custom task.

Udiknedormin: Oh, I see! Ok, so I'll factor these conditions out and add looking for `'apiVersion'` --- regular tasks don't have it while custom tasks always do.

sdk/python/tests/compiler/testdata/custom_task_params.py
```diff
@@ -0,0 +1,92 @@
...
+        }
+    ]
+}
+
+from kfp_tekton.tekton import TEKTON_CUSTOM_TASK_IMAGES
+TEKTON_CUSTOM_TASK_IMAGES = TEKTON_CUSTOM_TASK_IMAGES.append(custom_task_image)
+
+
+def custom_task(resource_label: str, foo: str, bar: Any, pi: float) -> dsl.ContainerOp:
+    task = dsl.ContainerOp(
```
Tomcli, line 61: FYI, when KFP switching to the V2 (IR) compiler next year, this dsl.ContainerOp will no longer work. So we are slowly migrating the ContainerOp DSL out of our samples and tests. https://github.com/kubeflow/kfp-tekton/issues/670

Since we are still in the middle of this migration, we don't have to change it for this PR. But we should be aware of it for future changes.

PR #691, opened 2021-08-07
remove ContainerOp from test cases - part 2

yhwang:
update the testcase accordingly.

related to #670

### DIFF
--- a/sdk/python/tests/compiler/testdata/long_param_name.py
+++ b/sdk/python/tests/compiler/testdata/long_param_name.py
@@ -12,35 +12,53 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from kfp import dsl
+from kfp import dsl, components
 from kfp_tekton.compiler import TektonCompiler as Compiler
 
 
 def gcs_download_op(url):
-    return dsl.ContainerOp(
-        name='GCS - Download' * 4,
-        image='google/cloud-sdk:279.0.0',
-        command=['sh', '-c'],
-        arguments=['gsutil cat $0 | tee $1', url, '/tmp/results.txt'],
-        file_outputs={
-            'data' * 10: '/tmp/results.txt',
-        }
-    )
+    return components.load_component_from_text("""
+      name: %s
+      description: download
+      inputs:
+        - {name: url, type: String}
+      outputs:
+        - {name: %s, type: String}
+      implementation:
+        container:
+          image: google/cloud-sdk:279.0.0
+          command:
+          - sh
+          - -c
+          args:
+          - |
+            gsutil cat $0 | tee $1
+          - {inputValue: url}
+          - {outputPath: %s}
+    """ % ('GCS - Download' * 4, 'data' * 10, 'data' * 10))(url=url)
 
 
-class PrintOp(dsl.ContainerOp):
-    def __init__(self, name, msg):
-        super(PrintOp, self).__init__(
-            name=name,
-            image='alpine:3.6',
-            command=['echo', msg])
+print_op = components.load_component_from_text("""
+name:
+""")
 
 
-@dsl.pipeline(name="Some very long name with lots of words in it. " +
-                   "It should be over 63 chars long in order to observe the problem.")
-def main_fn(url1='gs://ml-pipeline-playground/shakespeare1.txt'):
+@dsl.pipeline(name="some-very-long-name-with-lots-of-words-in-it-" +
+                   "it-should-be-over-63-chars-long-in-order-to-observe-the-problem")
+def main_fn(url1: str = 'gs://ml-pipeline-playground/shakespeare1.txt'):
     download1_task = gcs_download_op(url1)
-    PrintOp('print' * 10, download1_task.output)
+    components.load_component_from_text("""
+      name: %s
+      description: print
+      inputs:
+        - {name: text, type: String}
+      implementation:
+        container:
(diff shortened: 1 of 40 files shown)


yhwang:
One specific error from v2 compiler:
- NotImplementedError: dsl.graph_component is not yet supported in KFP v2 compiler.

The following files use `dsl.graph_component`:
- long_recursive_group_name.py
- loop_in_recursion.py
- nested_recur_custom_task.py
- nested_recur_runafter.py

And `old_kfp_volume.py` seems needs to interact with ml-pipeline api. I skipped this file for now.

pugangxa:
> One specific error from v2 compiler:
> 
> * NotImplementedError: dsl.graph_component is not yet supported in KFP v2 compiler.
> 
> The following files use `dsl.graph_component`:
> 
> * long_recursive_group_name.py
> * loop_in_recursion.py
> * nested_recur_custom_task.py
> * nested_recur_runafter.py

I saw graph_component in the change set, so we still do it for v1 and fix errors for v2 in future right?

yhwang:
> I saw graph_component in the change set, so we still do it for v1 and fix errors for v2 in future right?

Yes, and assuming it will be implemented in v2 compiler.

sdk/python/tests/compiler/testdata/input_artifact_raw_value.py
```diff
@@ -14,18 +14,28 @@
...
+    - {inputValue: text}
+"""
+comp_op = components.load_component_from_text(COMP_STR)
+
 
 def component_with_inline_input_artifact(text: str):
-    return dsl.ContainerOp(
-        name='component_with_inline_input_artifact',
-        image='alpine',
-        command=['cat', dsl.InputArgumentPath(text, path='/tmp/inputs/text/data', input='text')],  # path and input are optional
```
Tomcli, line 27: In this example, we want to test the `dsl.InputArgumentPath` DSL, so we can't simply just put text as argument.

sdk/python/tests/compiler/testdata/long_param_name.py
```diff
@@ -12,35 +12,53 @@
...
+          image: google/cloud-sdk:279.0.0
+          command:
+          - sh
+          - -c
+          args:
+          - |
+            gsutil cat $0 | tee $1
+          - {inputValue: url}
+          - {outputPath: %s}
+    """ % ('GCS - Download' * 4, 'data' * 10, 'data' * 10))(url=url)
```
Tomcli, line 38: can we try to create a long file name, since the output name is based on this.

```suggestion
    """ % ('GCS - Download' * 4, 'data' * 100, 'data' * 100))(url=url)
```

sdk/python/tests/compiler/testdata/loop_static.py
```diff
@@ -25,29 +25,61 @@ def empty(self):
...
+          - {name: input1, type: Integer}
+          - {name: input2, type: String}
+        implementation:
+          container:
+            image: library/bash:4.4.23
+            command:
+            - sh
+            - -c
+            args:
+            - |
```
Tomcli, line 44: You don't need to run the `set -x` command here?

sdk/python/tests/compiler/testdata/input_artifact_raw_value.py
yhwang: then need to skip this file and update it when we figure out how to use component.xml with dsl.InputArgumentPath

sdk/python/tests/compiler/testdata/loop_static.py
yhwang: original code doesn't have `set -x`. you want me to add `set -x` here?

sdk/python/tests/compiler/testdata/long_param_name.py
yhwang: then it throws error

sdk/python/tests/compiler/testdata/input_artifact_raw_value.py
yhwang: reverted it and I have a comment at issue [670](https://github.com/kubeflow/kfp-tekton/issues/670#issuecomment-895371381)

sdk/python/tests/compiler/testdata/loop_static.py
Tomcli: Oh i just wondering is the script still working without it, since I saw you added it in a lot of the examples in part 1 https://github.com/kubeflow/kfp-tekton/pull/685/files#diff-3b0c57e8f8c434d4d1e5338976ca3a0f5b41545519de957ffb53b84db38ba74eR59

sdk/python/tests/compiler/testdata/long_param_name.py
Tomcli: oh ok then is good

Tomcli: we don't want to have long name that we can't catch in our compiler

sdk/python/tests/compiler/testdata/loop_static.py
yhwang: actually, the key part is `sh -c`. I don't add `set -e` for the script part now. want to eliminate the diff in the yaml files.

Tomcli: cool thanks

PR #693, opened 2021-08-09
remove ContainerOp from test cases - part 3

yhwang:
update the testcase accordingly, including filenames starting
with `o` to `u`

related to #670

### DIFF
--- a/sdk/python/tests/compiler/testdata/pipelineparams.yaml
+++ b/sdk/python/tests/compiler/testdata/pipelineparams.yaml
@@ -15,7 +15,7 @@
 apiVersion: tekton.dev/v1beta1
 kind: PipelineRun
 metadata:
-  name: pipelineparams
+  name: pipeline-params
   annotations:
     tekton.dev/output_artifacts: '{"download": [{"key": "artifacts/$PIPELINERUN/download/downloaded_resultOutput.tgz",
       "name": "download-downloaded_resultOutput", "path": "/tmp/results.txt"}]}'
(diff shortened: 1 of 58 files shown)


sdk/python/tests/compiler/testdata/pipelineparams.yaml
```diff
@@ -45,48 +45,43 @@ spec:
...
-        value: $(params.sleep_ms)
-      - name: tag
-        value: $(params.tag)
       taskSpec:
         steps:
         - name: main
           args:
-          - sleep $(inputs.params.sleep_ms); wget localhost:5678 -O $(results.downloaded-resultoutput.path)
+          - |
+            sleep None; wget localhost:5678 -O $0
```
Tomcli, line 53: it should be `sleep $(inputs.params.sleep_ms);`. Do you think valuable replacement for image works in component.yaml?

sdk/python/tests/compiler/testdata/retry.py
```diff
@@ -12,23 +12,33 @@
...
+      - {name: exitcode, type: String}
+    implementation:
+      container:
+        image: python:alpine3.6
+        command:
+        - python
+        - -c
+        args:
+        - |
+          import random; import sys; \
```
Tomcli, line 33: do you think we need a new line here? when compiling it, the newline makes the spacing look very bad.

sdk/python/tests/compiler/testdata/sequential_noninlined.yaml
```diff
@@ -74,7 +85,7 @@ spec:
         steps:
         - name: main
           args:
-          - echo "$0"
+          - echo
```
Tomcli, line 88: quick question, does this command still work without "$0"?

yhwang: yes, for the old one, it uses parameter substitution. the new one just two arguments. both produce the same output

sdk/python/tests/compiler/testdata/pipelineparams.yaml
yhwang: I guess it doesn't work. then we need to fix the compiler error. let me revert this and put a comment on issue 670

sdk/python/tests/compiler/testdata/retry.py
yhwang: sure, merge it with next line

PR #702, opened 2021-08-17
Make results to store in multi-steps if needed

### LINKED ISSUE #701: Separate steps for writing results in case of multiple results
/kind feature

**Problem:**
When multiple results are generated by a `ContainerOp`, Tekton cannot get their values correctly due to utilising step's termination message and it having a hard limit of 4KB (in practice closer to 3KB).

Here is an example of how it fails for as few as seven "parameter" results:
```
message: '[{"key":"models","value":"[\"abcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTUVWXYabcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTU\", \"abcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTUVWXYabcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTU\", \"abcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTUVWXYabcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTU\"]","resourceRef":{},"type":"TaskRunResult"},{"key":"winning-model","value":"abcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTUVWXYabcdefghijklmnopqrstuvwxyABCDEFGHIJKLMNOPQRSTU","resourceRef":{},"type":"TaskRunResult"},{"key":"model-metrics","value":"[{\"abcdefghijklmnop\":3.141592653589793,\"abcdefghijklmnopqrstuvwxy\":3.141592653589793,\"abcdefghijklmnopqrstuvwx0\":3.141592653589793,\"abcdefghij\":3.141592653589793,\"abcdefghijklmnopqrst\":3.141592653589793,\"abcdefghijklmnopq\":3.14159,\"abcdefghijklmn\":3.141592653589793,\"abcdefghijklmno\":3.141592653589793,\"abcdefghijklmnop0\":3.141592653589793,\"abcdefghijklmnopqrstuvwxyA\":3.141592653589793,\"abcdefghijklmnopqrstuvwxy0\":3.141592653589793,\"abcdefghijk\":3.141592653589793,\"abcdefghijklmnopqrstu\":3.141592653589793,\"abcdefghijklmnopqr\":3.141592653589793,\"abcdefghijklmn0\":3.141592653589793,\"abcdefghijklmno0\":3.141592653589793},{\"abcdefghijklmnop\":3.141592653589793,\"abcdefghijklmnopqrstuvwxy\":3.141592653589793,\"abcdefghijklmnopqrstuvwx0\":3.141592653589793,\"abcdefghij\":3.141592653589793,\"abcdefghijklmnopqrst\":3.141592653589793,\"abcdefghijklmnopq\":3.14159,\"abcdefghijklmn\":3.141592653589793,\"abcdefghijklmno\":3.141592653589793,\"abcdefghijklmnop0\":3.141592653589793,\"abcdefghijklmnopqrstuvwxyA\":3.141592653589793,\"abcdefghijklmnopqrstuvwxy0\":3.141592653589793,\"abcdefghijk\":3.141592653589793,\"abcdefghijklmnopqrstu\":3.141592653589793,\"abcdefghijklmnopqr\":3.141592653589793,\"abcdefghijklmn0\":3.141592653589793,\"abcdefghijklmno0\":3.141592653589793},{\"abcdefghijklmnop\":3.141592653589793,\"abcdefghijklmnopqrstuvwxy\":3.141592653589793,\"abcdefghijklmnopqrstuvwx0\":3.141592653589793,\"abcdefghij\":3.141592653589793,\"abcdefghijklmnopqrst\":3.141592653589793,\"abcdefghijklmnopq\":3.14159,\"abcdefghijklmn\":3.141592653589793,\"abcdefghijklmno\":3.141592653589793,\"abcdefghijklmnop0\":3.141592653589793,\"abcdefghijklmnopqrstuvwxyA\":3.141592653589793,\"abcdefghijklmnopqrstuvwxy0\":3.141592653589793,\"abcdefghijk\":3.141592653589793,\"abcdefghijklmnopqrstu\":3.141592653589793,\"abcdefghijklmnopqr\":3.141592653589793,\"abcdefghijklmn0\":3.141592653589793,\"abcdefghijklmno0\":3.141592653589793}]","resourceRef":{},"type":"TaskRunResult"},{"key":"winning-model-metric","value":"3.1415926535897930","resourceRef":{},"type":"TaskRunResult"},{"key":"optimized-metric","value":"abcdefghijklmno","resourceRef":{},"type":"TaskRunResult"},{"key":"status","value":"Completed","resourceRef":{},"type":"TaskRunResult"},{"key":"status-message","value":"Training completed successfully","resourceRef":{},"type":"TaskRunResult"},{"key":"StartedAt","value":"YYYY-MM-DDThh:mm:ss.nnnZ","reso'
```

It happens despite the fact that each result is generally reasonably big:
* models: 300 B
* winning-model: 96 B
* model-metrics: 1861 B
* winning-model-metric: 18 B
* optimized-metric: 15 B
* status: 9 B
* status-message: 31 B

Plus StartedAt etc. takes some space too.

Aside from result number 2, with 1861 B, all others seem very small (note how `models` is a list of only a few elements). This problem isn't unique to tasks with one big result, however, it may be reproduced with all results being very small, provided their number will be sufficiently big.

**Solution:**
The above happens due to he hard limit on step termination message. Because of that, it could be avoided by delegating some results to separate steps.

What I'd suggest is two-steps:
* pod annotation `result_max_size_estimation` --- kfp-tekton should split results into separate steps if that value multiplied by the number of results in some step is more than, say, 2 KB (should be tested if that's enough, can be customisable via some compiler config field)
* results getting an annotation (accessible to the compiler via V2 `.component_spec`) `max_size_estimation` --- kfp-tekton should overwrite that result's estimation by this value

Neither value can be larger than 2 KB.

Now, kfp-tekton should put the "parameter" results in blocks of at most 2 KB, starting from the smallest:

| step | size   | results              |
|------|-------:|:---------------------|
|   1  |   9 B  | status               |
|   1  |  15 B  | optimized-metric     |
|   1  |  18 B  | winning-model-metric |
|   1  |  31 B  | status-message       |
|   1  |  96 B  | winning-model        |
|   1  | 300 B  | models               |
|   2  | 1861 B | model-metrics        |

Another example:

| step | size   |
|------|-------:|
|   1  |   20 B |
|   1  |   50 B |
|   1  |  350 B |
|   1  |  500 B |
|   1  |  600 B |
|   2  |  600 B |
|   2  | 1200 B |
|   3  | 1800 B |

If there is any more than one step, the task should get an `emptyDir` workspace (or volume+volumeMount), `result-splitter`, e.g:
```yaml
name: kfp-tekton-result-splitter
path: /kfp-tekton/results
```

For each outputPath of results in other steps, they should be replaced with paths in `result-splitter` (`outputPath: abc-def` --> `$(kfp-tekton-result-splitter.path)/abc-def`), or `/tekton/results/abc-def` to `/kfp-tekton/results/abc-def` if it's executed after the current behaviour.

Next, a new step with a very minimal image containing `cp` should be added and execute `cp` from those paths to the respective results' Tekton paths:
```bash
cp /kfp-tekton/results/abc-def /tekton/results/abc-def
```

Of course it all only affects results that are to be passed as parameters, not the ones to be passed only as artefacts.

**Additional information:**
Considering how in the original Argo KFP, whenever some result is parameter or artefact depends on how it's used downstream, a problem for the end results may arise: what if some ContainerOp produces a large result that isn't consumed downstream at all? Is there any possibility to "force" a result into being treated as an artefact (e.g. if knowing it's simply too big)?

It may be a good idea to do that with results that are declared as expected to be big (e.g. 1 KB or more).

Tomcli:
Resolves #701 

When users have too many Tekton results that add up to more than 4KB, Tekton cannot store all of the results from the same container to the etcd because it's using the Kubernetes termination message to aggregate the result files. Therefore, we need to create extra steps to move the result files to Tekton home in order to work around this issue. This solution is suggested by the [Tekton document](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results) when a task has a lot of results.

Usage:
On a containerOp/component.yaml, add the estimated size for each result outputs in the annotations:
```
    output_estimation_json = {'param1': 1500, 'param2': 700, 'param3': 500, 'param4': 900}
    print_task = print_op().add_pod_annotation('tekton-result-sizes', json.dumps(output_estimation_json))
``` 

The estimated size is a JSON with key as the result name and the value is the size of the result in Bytes. The goal here is to not have more than estimated 2KB of results in the same Tekton step. When adding this JSON to the annotation, we need to dump the JSON into string since K8S annotations doesn't take JSON format as it. 

This is a POC to test the can this feature handles many results. I will improve the bin packing algorithm and add more concrete usage if we are satisfy with this approach. 

**Environment tested:**

* Python Version (use `python --version`): 3.8

**Checklist:**


Makefile
```diff
@@ -87,7 +87,7 @@ lint: venv ## Check Python code style compliance
 	@which flake8 > /dev/null || pip install flake8
 	@flake8 sdk/python --show-source --statistics \
 		--select=E9,E2,E3,E5,F63,F7,F82,F4,F841,W291,W292 \
-		--per-file-ignores sdk/python/tests/compiler/testdata/*:F841 \
+		--per-file-ignores sdk/python/tests/compiler/testdata/*:F841,F821 \
```
Tomcli, line 90: Flake8 has some issue to detect the `NamedTuple` type name. Since F821 is only about Python style, we can ignore it in our testdata only. 
https://stackoverflow.com/questions/37840142/how-to-avoid-flake8s-f821-undefined-name-when-has-been-installed-by-get

Tomcli:
/hold
cc @Udiknedormin

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -505,4 +532,59 @@ def _op_to_template(op: BaseOp,
...
+        step_bins = {0: 0}
+        step_counter = 0
+        # Group result files to not exceed max_byte_size
+        # TODO: group result files with better bin packing algorithm
+        for key, value in result_size_map.items():
+            try:
+                value = int(value)
+            except ValueError:
+                raise("Estimated value for result %s is %s, but it needs to be an integer." % (key, value))
+            if key in op_result_names:
```
Udiknedormin, line 560: No warning or anything if there is no such result --- it's a bit misleading. Between KFP, Argo, Tekton etc, `-` and `_` are often used interchangeably, so it could be surprising if it simply doesn't work when one is used instead of the other.

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -505,4 +532,59 @@ def _op_to_template(op: BaseOp,
...
+            raise("tekton-result-sizes annotation is not a valid JSON")
+        # Sort key orders based on values
+        result_size_map = dict(sorted(result_size_map.items(), key=lambda item: item[1]))
+        max_byte_size = 2048
+        verified_result_size_map = {0: {}}
+        op_result_names = [name['name'] for name in template['spec']['results']]
+        step_bins = {0: 0}
+        step_counter = 0
+        # Group result files to not exceed max_byte_size
+        # TODO: group result files with better bin packing algorithm
```
Udiknedormin, line 554: I don't think it matters that much, but as for a better algorithm, how about the following:
1. sort the results from biggest to smallest
2. while some results are left to process:
   1. find the biggest result that still fits into the current step
   2. if no result fits, start a new segment
   3. if some result fits, put it into the current segment
3. reverse the steps' order

That could handle situations like:
* 100
* 200
* 400
* 1500
* 1600

currently:

| step | result |
|:----:|-------:|
|      |        |
|   2  |   1600 |
|      |        |
|   1  |   1500 |
|      |        |
|   0  |    400 |
|   0  |    200 |
|   0  |    100 |


new:

| step | result |
|:----:|-------:|
|      |        |
|   1  |   1600 |
|   1  |    400 |
|      |        |
|   0  |   1500 |
|   0  |    200 |
|   0  |    100 |

sdk/python/kfp_tekton/compiler/_op_to_template.py
```diff
@@ -49,6 +50,32 @@ def _get_base_step(name: str):
...
+    Args:
+        step_number {int}: step number
+        result_maps {list}: list of maps bucketed with the result groups
+
+    Returns:
+        Dict[Text, Any]
+    """
+    args = [""]
+    for key in result_maps[step_number].keys():
+        args[0] += "mv %s%s $(results.%s.path); " % (TEKTON_HOME_RESULT_PATH, key, key)
```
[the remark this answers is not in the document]
Tomcli: done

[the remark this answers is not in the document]
Tomcli: I kept the ; so it will still run if the string is concat by other services.

sdk/python/kfp_tekton/compiler/_op_to_template.py
Tomcli: done, adding new warnings and test cases

sdk/python/kfp_tekton/compiler/_op_to_template.py
Tomcli: Updated the algorithm to add results from large to small, should be optimized as above.

Udiknedormin (approved):

Tomcli:
@ckadner @yhwang can one of you lgtm this PR? thanks

PR #723, opened 2021-08-31
Remove Argo template UI reference to avoid UX conflicts with v2 compatible mode

Tomcli:
Resolves #

There are some new logics to check for Argo yaml when the pipeline could be in v2 compatible mode. Remove the Argo condition so if V2 and V2 compatible mode are not valid, it fails back to render with our Tekton DAG library.

**Environment tested:**

**Checklist:**

### DIFF
--- a/frontend/src/pages/PipelineDetails.tsx
+++ b/frontend/src/pages/PipelineDetails.tsx
@@ -304,16 +304,21 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     if (templateString) {
       try {
         const template = JsYaml.safeLoad(templateString);
-        if (isFeatureEnabled(FeatureKey.V2)) {
-          const pipelineSpec = WorkflowUtils.convertJsonToV2PipelineSpec(templateString);
-          graphV2 = convertFlowElements(pipelineSpec);
-        } else {
+        if (WorkflowUtils.isTektonPipelineRunTemplate(template)) {
           graph = StaticGraphParser.createGraph(template!);
 
           reducedGraph = graph ? transitiveReduction(graph) : undefined;
           if (graph && reducedGraph && compareGraphEdges(graph, reducedGraph)) {
             reducedGraph = undefined; // disable reduction switch
           }
+        } else if (isFeatureEnabled(FeatureKey.V2)) {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/PipelineDetails.tsx
```diff
@@ -304,21 +304,16 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     if (templateString) {
       try {
         const template = JsYaml.safeLoad(templateString);
-        if (WorkflowUtils.isArgoWorkflowTemplate(template)) {
```
yhwang, line 307: can we have `isTektonPipelineRunTemplate` instead and checking the `kind` and `apiVersion` accordingly? then we can also throw error if the yaml is not pipelinerun crd?

Tomcli: Unfortunately, we don't have the Tekton template library in Javascript because the Tekton spec is still dramatically changing on some major releases. We are devoting the spec check to the API which can help us better catch the errors.

yhwang: understood, I checked the code and since the function `isArgoWorkflowTemplate` doesn't enforce schema check. It only check two fields - `kind` and `apiVersion`. these two are common for kube CRD. hence no tekton js lib is needed.
code like this works well:
```
export function isTektonPipelineRunTemplate(template: Workflow): boolean {
  if (template?.kind === 'PipelineRun' && template?.apiVersion?.startsWith('tekton.dev/')) {
    return true;
  }
  return false;
}
```
this could minimize the change in `PipelineDetails.tsx` and the utility function goes to `WorkflowUtils.ts`

similar to my change here: https://github.com/yhwang/kfp-tekton/commit/588ec637ab51d54a98a39e99bc3154acc500a73d

but your change is good too :-)

yhwang (approved):

frontend/src/pages/PipelineDetails.tsx
Tomcli: sure, I will make the kind comparison to not be case sensitive to minimize possible future changes.

Tomcli: actually I will use your changes since that's the same as Argo implementation.

yhwang:
verified it on multi-user deployment. works well! Thanks!

PR #736, opened 2021-09-22
Optimize kfp-tekon SDK compile time

### LINKED ISSUE #735: Optimize kfp-tekton SDK compile time
/kind feature

**Description:**
When compiling the flip coin example, it takes around 0.25s to compile the pipeline DSL to Tekton yaml which is not very ideal for our internal users. So we want to optimize the compile time as much as possible.

On the initial testing, most of the bottlenecks are coming from loading/writing the dictionary into yaml/json. Thus we want to reduce those steps as much as possible 

**Additional information:**

Tomcli:
Resolves #735 

Reduce the number of writing and loading into YAML/JSON during compile time. This should reduce the compile time for 0.08-0.1s. 

E.g. The flip-coin example is reduced from 0.25s to 0.13s. 

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_tekton_handler.py
+++ b/sdk/python/kfp_tekton/compiler/_tekton_handler.py
@@ -54,6 +54,7 @@ def _process_argo_vars(workflow):
         workflow = re.sub(regex_rule['argo_rule'], regex_rule['tekton_rule'], workflow)
     return workflow
 
+
 def _handle_tekton_pipeline_variables(pipeline_run):
     """
     Handle tekton pipeline level variables, such as context.pipelineRun.name
--- a/sdk/python/kfp_tekton/compiler/_tekton_handler.py
+++ b/sdk/python/kfp_tekton/compiler/_tekton_handler.py
@@ -18,6 +18,42 @@ from kfp_tekton.compiler._k8s_helper import sanitize_k8s_name
 from kfp_tekton.tekton import LOOP_GROUP_NAME_LENGTH
 
 
+def _process_argo_vars(workflow):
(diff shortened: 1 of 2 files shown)


Tomcli:
cc @Udiknedormin

Udiknedormin:
@Tomcli My benchmark shows that a minimal example compiles TWICE as fast now --- more than enough of an optimisation as for now.

Udiknedormin:
@Tomcli

One thing comes to mind: I noticed that `_write_workflow` has the ability to simply return the yaml in-memory:
```python
if package_path is None:
  return yaml_text
```
However, `_create_and_write_workflow` doesn't propagate that, as it doesn't return the `_write_workflow`'s result. In the original KFP it wasn't a real issue, as the implementation of `_create_and_write_workflow` was as simple as:
```python
    pipeline_func,
    pipeline_name,
    pipeline_description,
    params_list,
    pipeline_conf)
self._write_workflow(workflow, package_path)
_validate_workflow(workflow)
```
so one could simply use `_create_workflow` if really needed. It's to as easy for kfp-tekton, `_create_and_write_workflow` being 60 lines.

If either `compile` with `package_path=None` would be supported, or another method that would return a yaml (or maybe a list of yamls, considering the possibility of multiple resources?) in-memory --- that would probably result in an even bigger speed-up.

Tomcli:
> ```python
> _write_workflow(
> ```

For the new method, I will create it in a separate PR that returns the workflow dictionary since the bottlenecks are happening in yaml.load/dump.

Tomcli:
@yhwang can you lgtm this PR? thanks.

PR #742, opened 2021-10-04
fix(sdk) orig_params being overwritten in custom-tasks

Udiknedormin:
In some rare cases (difficult to reproduce), pipeline params in custom tasks are not recognised as `orig_params` and therefore `$(inputs.params.foo-bar)` is produced instead of `$(params.foo-bar)`.

When custom task spec is overwritten from its `taskSpec`, all fields were pasted as-is. However, one of those was `"orig_params": task_ref["params"]` --- which works differently the first and the second time it's called, thus dropping the actual information on original parameters.

This PR factors out that variable and pastes it in both places without dropping its previous value.

**Environment tested:**

* Python Version (use `python --version`): 3.9.0
* Tekton Version (use `tkn version`): irrelevant
* Kubernetes Version (use `kubectl version`): irrelevant
* OS (e.g. from `/etc/os-release`): irrelevant

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/compiler/compiler.py
+++ b/sdk/python/kfp_tekton/compiler/compiler.py
@@ -751,17 +751,19 @@ class TektonCompiler(Compiler):
             for key, value in custom_task_args.items():
               if key not in non_param_keys:
                 task_params.append({'name': key, 'value': value})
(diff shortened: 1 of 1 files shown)


Tomcli (approved):
/lgtm
/approve

PR #743, opened 2021-10-04
[test] fix variable overriding

yhwang:
rename the variable in the script to avoid using the variable in
the build.properties
also add more time for waiting the kfp-tekton deployment


yhwang:
verified the change at the toolchain build.

PR #749, opened 2021-10-09
Add doc for break feature for loop

pugangxa:
The feature to support break in pipelineloop is already implemented. This is to add the document for that.

Can refer to the code https://github.com/kubeflow/kfp-tekton/blob/c835d1470bd9e225b13c67e9b7dcc5ccdc20efb7/tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go#L495-L514
I just run the example https://github.com/kubeflow/kfp-tekton/blob/master/tekton-catalog/pipeline-loops/examples/pipelinespec-with-run-condition.yaml and it works well, you can see the loop is defined to run 3 times but break when condition met while just run 2 times.
```
➜  examples git:(master) ✗ k describe run simplepipelinelooprun
Name:         simplepipelinelooprun
Namespace:    gang-test
Labels:       custom.tekton.dev/pipelineLoop=conditionloop
              last-loop-task=task-fail
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"custom.tekton.dev/v1alpha1","kind":"PipelineLoop","metadata":{"annotations":{},"name":"conditionloop","namespace":"gang-tes...
API Version:  tekton.dev/v1alpha1
Kind:         Run
......................
Events:
  Type    Reason     Age                    From              Message
  ----    ------     ----                   ----              -------
  Normal  Started    2m36s (x2 over 2m36s)  run-pipelineloop
  Normal  Running    2m36s (x2 over 2m36s)  run-pipelineloop  Iterations completed: 0
  Normal  Running    2m10s                  run-pipelineloop  Iterations completed: 1
  Normal  Succeeded  2m5s (x2 over 2m5s)    run-pipelineloop  PipelineRuns completed successfully with the conditions are met
```

### DIFF
--- a/tekton-catalog/pipeline-loops/README.md
+++ b/tekton-catalog/pipeline-loops/README.md
@@ -35,7 +35,7 @@
 `Parallelism` is to define the number of pipelineruns can be created at the same time. It must be eqaul to or bigger than 1. If it's not set then the default value will be 1. If it's bigger than the total iterations in the loop then the number will be total iterations.
 
 # Break
-It's common to break the loop when some condition met, like what we do in program language. This can be done by specify the task name with the `last-loop-task` label. When the task speicfied the the label is skipped during pipelineloop iteration, the loop will be marked `Succeeded` with a `pass` condition, and no more iteration will be started. The common use case is to check the condition in the task with `when` expression so that to decide whether to break the loop or just continue.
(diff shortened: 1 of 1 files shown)


pugangxa:
@Udiknedormin I have added the document for break feature for loop, please let me know if something unclear.
Note:
- Now there's no DSL interface in `dsl.ParallelFor`, may need enhance.
- Only one task can be specified now, may need enhance. Before remember we considered to specify it in CRD and support list but seems it's enough with current design.
- For parallelism it will just break and stop starting new pipelineruns but the ones that already started will continue running.

pugangxa:
> @pugangxa @Udiknedormin is this PR ready to merge?
> 
> For DSL support on this feature, we probably need to know some workflow use cases. If we just want to experiment with it, we can use the task annotations as the feature flag to enable breaks.

Yes there's requirements to support break in loop and this is to show the ability of that in pipeline loop now, I think we can let it go currently and enhance later when requirement is more clear, thanks.

PR #757, opened 2021-10-20
feat(sdk): Add an API to return prepared workflow as python dicts.

### LINKED ISSUE #737: Create new compiler helper method to return workflow dictionary only
@Tomcli

One thing comes to mind: I noticed that `_write_workflow` has the ability to simply return the yaml in-memory:
```python
if package_path is None:
  return yaml_text
```
However, `_create_and_write_workflow` doesn't propagate that, as it doesn't return the `_write_workflow`'s result. In the original KFP it wasn't a real issue, as the implementation of `_create_and_write_workflow` was as simple as:
```python
    pipeline_func,
    pipeline_name,
    pipeline_description,
    params_list,
    pipeline_conf)
self._write_workflow(workflow, package_path)
_validate_workflow(workflow)
```
so one could simply use `_create_workflow` if really needed. It's to as easy for kfp-tekton, `_create_and_write_workflow` being 60 lines.

Considering how `compile` is based on Argo KFP's version, it probably cannot be changed that way, but maybe another method that would return a yaml (or maybe a list of yamls, considering the possibility of multiple resources?) in-memory could be added? That could possibly result in an even bigger speed-up for some cases.

_Originally posted by @Udiknedormin in https://github.com/kubeflow/kfp-tekton/issues/736#issuecomment-925200043_

We want to have a new method to return the workflow dictionary in Python memory to optimize the API response time when running the SDK as a service.

ScrapCodes:
Resolves #737

Add an API (i.e. prepare_workflow) to return prepared workflow as python dicts.

Advantage of this API is, that is saves few microseconds spent in compiling python dict in yaml and store on disk, 
because it returns the python dict in memory.

Disadvantage of the approach,

It will allow the end users to direct consume internal data-structure and they may tend to rely on its structure. In future, we do not guarantee that the fields or structure of field will not change in this dict.

**Environment tested:**

**Checklist:**

### DIFF
--- a/guides/kfp-user-guide/README.md
+++ b/guides/kfp-user-guide/README.md
@@ -59,9 +59,58 @@ from kfp_tekton.compiler import TektonCompiler
 TektonCompiler().compile(echo_pipeline, 'echo_pipeline.yaml')
 ```
 
-In addition, we can put the above python code into a python `__main__` function. Then, the compilation can be done in terminal with a simple python command.
+Place the above python code into a python `__main__` function. Then, the compilation can be done in terminal with a simple python command.
 ```shell
 python echo_pipeline.py
+```
+
+Additionally, if user wants to have a python dict of compiled output for further processing. Use `prepare_workflow` API.
+
+```python
+from kfp_tekton.compiler import TektonCompiler
+python_dict = TektonCompiler().prepare_workflow(echo_pipeline)
+
+pprint(python_dict)
+```
+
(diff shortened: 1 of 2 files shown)


Tomcli:
thanks @ScrapCodes, can you add a simple example on how to use this new function? You can just extend a new section in the kfp-user-guide over here https://github.com/kubeflow/kfp-tekton/tree/master/guides/kfp-user-guide

PR #773, opened 2021-11-17
Add performance test script

ckadner:

Adding a rudimentary performance test script, to be improved over time.

The script loads a small set of pipelines, currently from the compiler testdata.

For each pipeline it times ...
* pipeline compilation
* pipeline run submit
* pipeline run time

And adds the records to a CSV file:

|Pipeline|Compile|Submit|Run|
|-|-|-|-|
|sequential_pipeline|0:00:00.081066|0:00:00.322569|0:00:29.895088|
|flipcoin|0:00:00.123511|0:00:00.646792|0:01:42.271202|
|save_most_frequent_word|0:00:00.042771|0:00:00.369670|0:01:29.414623|
|retry_sample_pipeline|0:00:00.072874|0:00:00.277650|0:00:40.157533|
|pipeline|0:00:00.103826|0:00:00.322516|0:00:26.575043|
|conditions_and_loops|0:00:00.144815|0:00:00.483953|0:02:06.636966|
|...| | | |

**Environment variables to set:**

```Python
PUBLIC_IP = env.get("PUBLIC_IP", "169....")
USER_INFO = env.get("USER_INFO", '{"username":"admin", ...}')
CONNECT_SID = env.get("CONNECT_SID", "s:OerT...m3A")
NAMESPACE = 
```

**How to run it:**

```Bash
export PUBLIC_IP="169...."

# in case of multi-user deployment, specify the following env vars
export NAMESPACE=...
export USER_INFO='{"username":"...","email":"...","roles":["admin"]}' 
export CONNECT_SID="s:Oer***************fRKNm3A" 

cd sdk/python/tests/compiler
python3 performance_tests.py 
# ... wait for script to finish
cat performance_test_times_*.csv
```

Running the test script with 10 concurrent workers:

```Bash
NUM_WORKERS=10 sdk/python/tests/compiler/performance_tests.py
```

**Caveats/TODOs:**

- [ ] add script parameters instead of ENV vars
- [ ] KFP client authentication requires a browser login to scrape the auth cookie (`connect.sid` and `userinfo`)
- [ ] add script documentation
- [ ] parameterize output file path instead of creating it in current working directory
- [ ] list of pipelines is incomplete
- [x] distinguish pipeline functions that have duplicated names
- [ ] pipelines with custom tasks don't work yet -- they crash the ml-pipeline service
- [x] run pipelines in parallel
- [ ] get execution times for each task

### DIFF
--- a/sdk/python/tests/compiler/performance_tests.py
+++ b/sdk/python/tests/compiler/performance_tests.py
@@ -35,12 +35,14 @@ from kfp_tekton.compiler.pipeline_utils import TektonPipelineConf
 # =============================================================================
 
 # TODO: remove defaults below, add instructions and/or script parameters
-PUBLIC_IP = env.get("PUBLIC_IP")  # mlx-demo
+PUBLIC_IP = env.get("PUBLIC_IP")
+NAMESPACE = env.get("NAMESPACE", None)
 USER_INFO = env.get("USER_INFO")
 CONNECT_SID = env.get("CONNECT_SID")
 
 print(f"Environment variables:\n"
       f"  PUBLIC_IP:   {PUBLIC_IP}\n"
+      f"  NAMESPACE:   {NAMESPACE}\n"
       f"  USER_INFO:   {USER_INFO}\n"
       f"  CONNECT_SID: {CONNECT_SID}\n")
 
@@ -56,8 +58,9 @@ experiment_name = "PERF_TEST"
 
 def get_client() -> TektonClient:
     host = f"http://{PUBLIC_IP}/pipeline"
-    cookies = f"connect.sid={CONNECT_SID}; userinfo={USER_INFO}"
+    cookies = f"connect.sid={CONNECT_SID}; userinfo={USER_INFO}" if CONNECT_SID and USER_INFO else None
     client = TektonClient(host=host, cookies=cookies)
+    client.set_user_namespace(NAMESPACE)  # overwrite system default with None if necessary
     return client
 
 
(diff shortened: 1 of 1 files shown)


Tomcli:
thanks @ckadner, for pipelines with custom task, you need to run with the custom task feature flag enabled. You can use the two performance cluster I have set up since that is configured to run both pipelineloop and cel custom tasks.

ckadner:
Thanks @Tomcli -- I will make a few more updates. I am currently testing a few more pipelines on one of your performance-dedicated cluster

sdk/python/tests/compiler/performance_tests.py
```diff
@@ -0,0 +1,260 @@
...
+    host = f"http://{PUBLIC_IP}/pipeline"
+    cookies = f"connect.sid={CONNECT_SID}; userinfo={USER_INFO}" if CONNECT_SID and USER_INFO else None
+    client = TektonClient(host=host, cookies=cookies)
+    client.set_user_namespace(NAMESPACE)  # overwrite system default with None if necessary
+    return client
+
+
+# method annotation to ensure the wrapped function is executed synchronously
+def synchronized(function):
+    lock = threading.Lock()
```
yhwang, line 79: could this lock be moved down to inside the wrapped function to even narrow down the locking scope? or you put it here to create the lock above the wrapped function on purposed

ckadner: the only function that needs to be synchronized, in the current implementation of this script is the pipeline compilation

ckadner: I will have to rewrite this script to record execution times differently in order to record compile time accurately, but the compilation times are below a 1/10 of a second so I don't think it matters to much

ckadner: I rewrote the test script such that compile times are recorded accurately now, despite having to synchronize access to the `compile_pipeline` method.

PR #776, opened 2021-11-19
expose current iteration item to pipelinerun's annotation

wzhanw:
Resolves #
We want to get the current iteration item in the pipeline's task node for condition check and any other purpose.
Any task run created by this pipeline will inherit the annotation list.  In our customer run we can get the item value and to our logic.

Get the current iteration item and add it to the pipelinerun's annotation list.

### DIFF
--- a/tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go
+++ b/tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go
@@ -233,7 +233,7 @@ func (c *Reconciler) reconcile(ctx context.Context, run *v1alpha1.Run, status *p
 	}
 
 	// Determine how many iterations of the Task will be done.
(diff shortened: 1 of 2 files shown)


tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go
```diff
@@ -230,7 +233,7 @@ func (c *Reconciler) reconcile(ctx context.Context, run *v1alpha1.Run, status *p
 	}
 
 	// Determine how many iterations of the Task will be done.
-	totalIterations, err := computeIterations(run, pipelineLoopSpec)
+	totalIterations, interationElements, err := computeIterations(run, pipelineLoopSpec)
```
pugangxa, line 236: There's many `interationElements`, I guess `interation` should be `iteration`?

wzhanw: you are right, I corrected the typo.

PR #795, opened 2021-12-01
Fix performance test script and update testing examples

Tomcli:
Resolves #

Fix the custom task bug in the performance test script and update the list of testing example to be sequential. 

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/tests/compiler/performance_tests.py
+++ b/sdk/python/tests/compiler/performance_tests.py
@@ -208,23 +208,38 @@ def get_client() -> TektonClient:
 def load_pipeline_functions() -> [(Callable, str)]:
     pipeline_functions = []
 
+    # Sequential pipelines
+    from testdata.condition_sample import flipcoin_pipeline as normal_flipcoin_pipeline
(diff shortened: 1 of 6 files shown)


sdk/python/tests/compiler/performance_tests.py
```diff
@@ -305,16 +319,16 @@ def get_details(data):
         rev["taskRuns"] = get_details(status["taskRuns"])
 
     if "runs" in status:
-        rev["run"] = get_details(status["runs"])
+        rev["runs"] = get_details(status["runs"])
```
ckadner, line 322: I also added this fix in my PR #796

ckadner (commented):
Funky timing :-)  I guess I took too long with my PR :-\

Did you copy those new pipelines from `samples` to `testdata`, or also make changes to them so that it makes sense to keep both copies? Also for compiler testing?

If the original files under sample are the same, then the changes from my PR might be enough?

Btw, I added your fix to `rev["run"] = get_details(status["runs"])`  in my PR as well. 

Do you still want the ability to load pipeline files from anywhere?

ckadner (approved):
/lgtm

Issue #791, opened 2021-11-26
feature(sdk): "Loop" DSL extension


Udiknedormin:
/kind feature

**Description:**
There are multiple features that `PipelineLoop` offers that are lacking in the original KFP's `dsl.ParallelFor`. The DSL users, therefore, cannot easily use them --- in order to do that they need to first generate yaml from DSL and then modify the resulting yaml accordingly, which isn't always trivial.

My proposal: to introduce a DSL extension for `Loop`. Usage:

```python
# just like dsl.ParallelLoop
with Loop(array, parallelism=5) as item:
  pass

# range, "step" being optional
with Loop(start=1, end=10) as item:
  pass

# with string separator
with Loop(string, separator=",") as item:
  pass

# with "break" capability (what is now last-loop-task)
with Loop(array) as item:
  with dsl.Condition(CEL_ConditionOp(f"{item} == 0")):
    Break()
```

To make the transition a bit easier, it can inherit from dsl.ParallelFor:
```python
_Num = TypeVar('_Num', int, float)
class Loop(dsl.ParallelFor):
  @classmethod
  def sequential(cls,
                 loop_args: _for_loop.ItemList):
    return cls(loop_args, parallelism=1)

  @classmethod
  def from_string(cls,
                  loop_args: Union[str, _pipeline_param.PipelineParam],
                  separator: Optional[str] = None,
                  **,
                  parallelism: Optional[int] = None):
    return cls(loop_args=string, separator=separator, parallelism=parallelism)

  @overload
  @classmethod
  def range(cls,
            start: _Num,
            end: _Num,
            **,
            parallelism: Optional[int] = None): ...
  @overload
  @classmethod
  def range(cls,
            start: _Num,
            step: _Num,
            end: _Num,
            **,
            parallelism: Optional[int] = None): ...

  @classmethod
  def range(cls,
            a: _Num,
            b: _Num,
            c: Optional[_Num] = None,
            **,
            parallelism: Optional[int] = None):
      if c is None:
          return cls(start=a, end=b, parallelism=parallelism)
      else:
          return cls(start=a, step=b, end=c, parallelism=parallelism)

  @overload
  def __init__(self,
               loop_args: _for_loop.ItemList,
               parallelism: Optional[int] = None): ...
  @overload
  def __init__(self,
               loop_args: _pipeline_param.PipelineParam,
               parallelism: Optional[int] = None,
               separator: Optional[str] = None): ...
  @overload
  def __init__(self,
[...]

PR #799, opened 2021-12-15
Add tekton loop dsl extension skeleton

Tomcli:
Related #791

Add SDK skeleton for Tekton specific DSL extensions.

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/tekton.py
+++ b/sdk/python/kfp_tekton/tekton.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Iterable, Union
+from typing import List, Iterable, Union, Optional, TypeVar
+from kfp.dsl import _pipeline_param, _for_loop
 from kfp import dsl
 from kfp import components
 from kfp.dsl._pipeline_param import ConditionOperator
 from kfp_tekton.compiler._k8s_helper import sanitize_k8s_name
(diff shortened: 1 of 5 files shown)


sdk/python/kfp_tekton/tekton.py
```diff
@@ -177,3 +180,68 @@ def CEL_ConditionOp(condition_statement):
...
+            - -c
+            - |
+              echo "$0"
+            args:
+            - "break loop"
+    ''' % (TEKTON_BASH_STEP_IMAGE)
+    BreakOp_template = components.load_component_from_text(BreakOp_yaml)
+    BreakOp = BreakOp_template()
+    return BreakOp
+
```
pugangxa, line 205: To be honest I did not follow how the break happens here. Seems it's just a normal task?

Udiknedormin: @pugangxa I think it's related to issue #800.

Tomcli: Yes, right now the pipeline loop break only works when the `last-loop-task` is skipped. However, the proposed `Break()` DSL is to break when the condition is met. So I opened #800 to add a new break features to break when the `Break()` task is being executed.

pugangxa: Sorry I just overlooked that issue. Then it makes sense. Thanks for all your explanation.

PR #802, opened 2021-12-15
Added log to object store.

ScrapCodes:
Resolves #

1. Added a new config map for supplying the object store configurations.
2. Added log to object store for every log message

TODO:
1. Next the logger itself should somehow implement the logging interface of `golang` so that it can log even internal libraries log messages to object store.
2. Add a buffered or batched logging, i.e. instead of logging every message to object store batch them at some time interval. This will improve performance by regulating the no. of calls to object store, as there is a performance penalty for a call to object store.

**Environment tested:**

**Checklist:**


tekton-catalog/pipeline-loops/config/203-object-store-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: ConfigMap
+metadata:
+  name: object-store-config
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines-loops
+data:
+  enable: "false"
+  apiKey: "<APIKEY-dummy-1231231231-123abcdefgh>"
```
Tomcli, line 25: Can we also have an option to push to object storage using HMAC S3 keys. (e.g. `access_key_id` and `secret_access_key`?) So we can use this option with minio storage for open source as well.

Tomcli:
Thanks @ScrapCodes. On the high level it looks good. As you mention, able to move this code as a separate package and push logs in batch are important for our performance optimization. 

For this PR, let's just fix the tests and have options to hook this up with minio as well. Then I can merge it and you can improve it in your follow up PRs.

ScrapCodes:
/hold

This PR is an alternate approach to #806. 
Advantage of this approach is, it allows for producing the logs in separate for each `Run`. 

But disadvantages are the 

1. Next the logger itself should somehow implement the logging interface of golang so that it can log even internal libraries log messages to object store.

2. Add a buffered or batched logging, i.e. instead of logging every message to object store batch them at some time interval. This will improve performance by regulating the no. of calls to object store, as there is a performance penalty for a call to object store.

PR #828, opened 2022-01-19
Additional param to enable running copies of same pipeline concurrently

tedhtchang:
Signed-off-by: ted chang <htchang@us.ibm.com>

Resolves # 780

Add additional param, `copies`, to enable running N copies of same pipeline concurrently.  

**Environment tested:**

* Python Version (use `python --version`): 3.9.6
* Tekton Version (use `tkn version`): v0.27.0
* Kubernetes Version (use `kubectl version`): v1.21.7
* OS (e.g. from `/etc/os-release`): OSX 12.1

### DIFF
--- a/sdk/python/tests/performance_tests.py
+++ b/sdk/python/tests/performance_tests.py
@@ -258,7 +258,7 @@ def load_pipeline_scripts() -> [(Path, str)]:
 
         path = path_name_dict["path"]
         name = path_name_dict.get("name") or Path(path).stem
-
+        copies = path_name_dict.get("copies", 1)
         if not path.startswith(pathsep):
             # path assumed to be relative to project root
             fp: Path = project_dir.joinpath(path)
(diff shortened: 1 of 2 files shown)


sdk/python/tests/performance_tests.py
```diff
@@ -258,7 +258,7 @@ def load_pipeline_scripts() -> [(Path, str)]:
 
         path = path_name_dict["path"]
         name = path_name_dict.get("name") or Path(path).stem
-
+        copies = path_name_dict.get("copies")
```
Tomcli, line 261: Maybe have a default value so it at least run once?
```suggestion
        copies = path_name_dict.get("copies", 1) 
```

sdk/python/tests/performance_tests.py
```diff
@@ -267,8 +267,8 @@ def load_pipeline_scripts() -> [(Path, str)]:
             fp = Path(path)
 
         assert fp.exists(), f"Cannot find file: {fp.resolve()}"
-
-        pipeline_files_with_name.append((fp, name))
+        for i in range(copies):
```
Tomcli, line 270: Just to check it can convert to int before use it as range.
```suggestion
        for i in range(int(copies)):
```

PR #838, opened 2022-01-26
Add loop numeric sdk support

### LINKED ISSUE #791: feature(sdk): "Loop" DSL extension

Tomcli:
Resolves #791 

Support basic numeric loop capability on the SDK.

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_tekton_handler.py
+++ b/sdk/python/kfp_tekton/compiler/_tekton_handler.py
@@ -233,6 +233,17 @@ def _handle_tekton_custom_task(custom_task: dict, workflow: dict, recursive_task
             separator = custom_task[custom_task_key].get('separator')
             if separator is not None:
                 custom_task_cr['spec']['iterateParamStringSeparator'] = separator
+            if custom_task[custom_task_key].get('start') is not None:
+                start_end_step_keys = ['from', 'to', 'step']
+                for i, value in enumerate(custom_task_cr['spec']['pipelineSpec']['params']):
+                    if value['name'] in start_end_step_keys:
+                        custom_task_cr['spec']['pipelineSpec']['params'].pop(i)
+                for i, custom_task_param in enumerate(custom_task[custom_task_key]['spec']['params']):
+                    if custom_task_param['name'] == custom_task_cr['spec']['iterateParam']:
+                        custom_task[custom_task_key]['spec']['params'].pop(i)
+
+                custom_task_cr['spec']['iterateNumeric'] = custom_task_cr['spec']['iterateParam']
+                custom_task_cr['spec'].pop('iterateParam')
             for custom_task_param in custom_task[custom_task_key]['spec']['params']:
                 if custom_task_param['name'] != custom_task[custom_task_key]['loop_args'] and '$(tasks.' in custom_task_param['value']:
                     custom_task_cr = json.loads(
@@ -352,6 +363,15 @@ def _handle_tekton_custom_task(custom_task: dict, workflow: dict, recursive_task
             if task['name'].replace(task_name_prefix, "") not in task_list_trimmed:
                 new_tasks.append(task)
     workflow['spec']['pipelineSpec']['tasks'] = new_tasks
+    # clean up any redundant parameters
+    for i, task in enumerate(workflow['spec']['pipelineSpec']['tasks']):
+        searched_params = []
+        if task.get('params'):
+            for j, workflow_param in enumerate(task['params']):
+                if workflow_param['name'] in searched_params:
+                    workflow['spec']['pipelineSpec']['tasks'][i]['params'].pop(j)
+                else:
+                    searched_params.append(workflow_param['name'])
     return custom_task_crs, workflow
 
 
--- a/sdk/python/kfp_tekton/compiler/_tekton_handler.py
+++ b/sdk/python/kfp_tekton/compiler/_tekton_handler.py
@@ -235,12 +235,12 @@ def _handle_tekton_custom_task(custom_task: dict, workflow: dict, recursive_task
                 custom_task_cr['spec']['iterateParamStringSeparator'] = separator
             if custom_task[custom_task_key].get('start') is not None:
                 start_end_step_keys = ['from', 'to', 'step']
-                for i, value in enumerate(custom_task_cr['spec']['pipelineSpec']['params']):
-                    if value['name'] in start_end_step_keys:
-                        custom_task_cr['spec']['pipelineSpec']['params'].pop(i)
-                for i, custom_task_param in enumerate(custom_task[custom_task_key]['spec']['params']):
-                    if custom_task_param['name'] == custom_task_cr['spec']['iterateParam']:
-                        custom_task[custom_task_key]['spec']['params'].pop(i)
+                custom_task_cr['spec']['pipelineSpec']['params'] = [value for value
+                                                                    in custom_task_cr['spec']['pipelineSpec']['params']
+                                                                    if value['name'] not in start_end_step_keys]
+                custom_task[custom_task_key]['spec']['params'] = [custom_task_param for custom_task_param
+                                                                  in custom_task[custom_task_key]['spec']['params']
+                                                                  if custom_task_param['name'] != custom_task_cr['spec']['iterateParam']]
 
                 custom_task_cr['spec']['iterateNumeric'] = custom_task_cr['spec']['iterateParam']
                 custom_task_cr['spec'].pop('iterateParam')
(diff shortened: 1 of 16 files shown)


Tomcli:
cc @Udiknedormin

sdk/python/kfp_tekton/compiler/_tekton_handler.py
```diff
@@ -352,6 +363,15 @@ def _handle_tekton_custom_task(custom_task: dict, workflow: dict, recursive_task
             if task['name'].replace(task_name_prefix, "") not in task_list_trimmed:
                 new_tasks.append(task)
     workflow['spec']['pipelineSpec']['tasks'] = new_tasks
+    # clean up any redundant parameters
+    for i, task in enumerate(workflow['spec']['pipelineSpec']['tasks']):
+        searched_params = []
+        if task.get('params'):
+            for j, workflow_param in enumerate(task['params']):
+                if workflow_param['name'] in searched_params:
+                    workflow['spec']['pipelineSpec']['tasks'][i]['params'].pop(j)
```
Udiknedormin, line 372: This line pops from an array that is being iterated. Would be better to either store the indexes to remove or storing the new array to assign in one go (and not appending to it in such a case).

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -435,6 +439,40 @@ def process_pipelineparam(s):
...
+      if hasattr(sub_group, 'start') and sub_group.start is not None:
+        # start, end, step params should be added as a parameter
+        # isinstance(sub_group.start, dsl.PipelineParam)
+        def process_parameter(parameter):
+          parameter_value = str(parameter)
+          if isinstance(parameter, dsl.PipelineParam):
+            if parameter.op_name:
+              parameter_value = '$(tasks.' + parameter.op_name + '.results.' + sanitize_k8s_name(parameter.name) + ')'
+            else:
+              parameter_value = '$(params.' + parameter.name + ')'
```
Udiknedormin, line 451: What if it's a combination of string and pipeline parameter or multiple pipeline parameters?

For instance:
```
start="{{pipelineparam:op=;name=integer}}.{{pipelineparam:op=;name=decimal}}"
```

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -435,6 +439,40 @@ def process_pipelineparam(s):
...
+          return parameter_value
+
+        def get_spec_param_keys(param_specs):
```
Udiknedormin, line 454: This function doesn't seem to be used...

sdk/python/kfp_tekton/tekton.py
```diff
@@ -259,30 +259,36 @@ def __init__(self,
...
-        super().__init__(loop_args=[loop_args], parallelism=parallelism)
-        self.loop_args = TektonLoopArguments(
-            loop_args,
-            code=next_id(),
-        )
-        self.items_is_string = True
+        super().__init__(loop_args=["iteration"], parallelism=parallelism)
+        self.start = start
+        self.end = end
+        self.step = step if step is not None else None
```
Udiknedormin, line 269: `step if step is not None else None` === `step`

sdk/python/tests/compiler/compiler_tests.py
```diff
@@ -126,6 +126,13 @@ def test_recur_cond_workflow(self):
     from .testdata.tekton_loop_dsl import pipeline
     self._test_pipeline_workflow(pipeline, 'tekton_loop_dsl.yaml')
 
+  def test_loop_with_numeric_workflow(self):
+    """
+    Test compiling a loop with literal separator in workflow.
```
Udiknedormin, line 131: *with numeric range

This test does covert start-end, but what about start-step-end scenario?

sdk/python/tests/compiler/testdata/loop_with_numeric.yaml
```diff
@@ -0,0 +1,157 @@
...
+                            pipelines.kubeflow.org/generation: ''
+                            pipelines.kubeflow.org/cache_enabled: "true"
+                          annotations:
+                            pipelines.kubeflow.org/component_spec_digest: '{"name":
+                              "my-in-coop1", "outputs": [], "version": "my-in-coop1@sha256=42f0c6a60c7e435441b8afaeb382a771a9741fe3aabb203748fdbd72b25f1628"}'
+                            tekton.dev/template: ''
+                      timeout: 525600m
+                  iterateNumeric: loop-item-param-3
+          iterateNumeric: loop-item-param-1
+  timeout: 525600m
```
Udiknedormin, line 157: Did you run it, by any chance?

Udiknedormin (commented):
An example with `step` would be nice too.

sdk/python/kfp_tekton/compiler/_tekton_handler.py
Tomcli: sure, I rewrote this into a new array instead of popping the items.

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: I think start, to, step only can loop over integers. Also, the start parameter can either be pipeline param or string, not both at the same time.

Tomcli: if this is necessary, we probably need to convert the pipeline param into string, then use some regular expression to parse the `{{pipelineparam:op=;name=integer}}` syntax back to Tekton params for this use case only. I think to make the compiler less complex, it's better to treat it like python range() where we should use these number as indices rather than decimal inputs.

sdk/python/kfp_tekton/compiler/compiler.py
Tomcli: I was thinking to use this function to get me all the parameter names to pass down to the nested loop. Since that feature needs some big changes to the controller, I removed it for now.

sdk/python/kfp_tekton/tekton.py
Tomcli: thanks, done

sdk/python/tests/compiler/testdata/loop_with_numeric.yaml
Tomcli: Yes I did. It works fine for this example.

sdk/python/tests/compiler/compiler_tests.py
Tomcli: I added one more example for start, end, step.

sdk/python/kfp_tekton/compiler/compiler.py
Udiknedormin: Sounds quite complex indeed. It's not a must for our team, just a quick thought that the general user might want to try that kind of thing.

I think what kfp-tekton could do right now is to try parse the pipelineparam from string. If there is only one and it matches the whole string --- then it's all good and fine. If not --- an error indicating that it's not supported can be raised.

yhwang (approved):
/lgtm

PR #840, opened 2022-01-27
Remove unsynced components

Tomcli:
Resolves #

Since KFP-Tekton and KFP shares the same components, we don't want to maintain an unsynced version of components as these components can be updated very frequently upstream. Thus, we removed all the components that are pulled from the upstream KFP. 

The notification component is an example of converting Tekton component to KFP component. Since it involves human translation to convert Tekton `workspace` and `resources` to KFP (since Tekton `resources` needs to create a new pod but KFP only needs an extra step), we would like to keep it as our component example for the record.

**Environment tested:**

**Checklist:**


yhwang (approved):
/lgtm

PR #866, opened 2022-03-01
Fix Tekton conditional dependency on sub-dag

### LINKED ISSUE #865: Loop on condition breaks the yaml
/kind bug

## What steps did you take and what happened:

When using a loop depending on a task inside of a condition:
```python
from kfp import dsl
from kfp.components import load_component_from_text
from kfp_tekton.tekton import CEL_ConditionOp


class Coder:
  def empty(self):
    return ""


TektonCompiler._get_unique_id_code = Coder.empty


class CelCondition(dsl.Condition):
  def __init__(self, pred: str, name: str = None):
    super().__init__(CEL_ConditionOp(pred).output == 'true', name)


def PrintOp(name: str, msg: str = None):
  if msg is None:
    msg = name
  print_op = load_component_from_text(
  """
  name: %s
  inputs:
  - {name: input_text, type: String, description: 'Represents an input parameter.'}
  outputs:
  - {name: output_value, type: String, description: 'Represents an output paramter.'}
  implementation:
    container:
      image: alpine:3.6
      command:
      - sh
      - -c
      - |
        set -e
        echo $0 > $1
      - {inputValue: input_text}
      - {outputPath: output_value}
  """ % (name)
  )
  return print_op(msg)


@dsl.pipeline("empty-loop")
def condition_1(param: list = ["a", "b", "c"]):
  op0 = PrintOp("print-0")

  with CelCondition(f'{op0.output} == "print-0"'):
    op1 = PrintOp("print-1")

  # works fine for a task outside of loop:
  op2 = PrintOp("print-2")
  op2.after(op1)

  # ...but breaks for loop or task inside of a loop:
  loop = dsl.ParallelFor(param)
  # both A) and B) yield the same result:
  # A)
  # loop.after(op1)
  with loop:
    op3 = PrintOp("print-3")
    # B)
    op3.after(op1)


if __name__ == '__main__':
  TektonCompiler().compile(condition_1, __file__.replace('.py', '.yaml'))
```

It yields:
```yaml
kind: PipelineRun
metadata:
  name: empty-loop
  annotations: (...)
spec:
  params:
  - name: param
    value: '["a", "b", "c"]'
  pipelineSpec:
    params:
    - name: param
      default: '["a", "b", "c"]'
    tasks:
    - name: print-0
      taskSpec: (...)
      timeout: 525600m
    - name: condition-cel
      params:
      - name: outcome
        value: $(tasks.print-0.results.output-value) == "print-0"
      taskRef:
        name: cel_condition
        apiVersion: cel.tekton.dev/v1alpha1
        kind: CEL
      timeout: 525600m
    - name: print-1
      taskSpec: (...)
      when:
      - input: $(tasks.condition-cel.results.outcome)
        operator: in
        values:
        - "true"
      timeout: 525600m
    - name: print-2
      taskSpec: (...)
      runAfter:
      - print-1
      timeout: 525600m
    - runAfter:
      - condition-1
      name: empty-loop-for-loop-2
      params:
      - name: param-loop-item
        value: $(params.param)
      taskSpec:
        apiVersion: custom.tekton.dev/v1alpha1
        kind: PipelineLoop
        spec:
          pipelineSpec:
            params:
            - name: param-loop-item
              type: string
            tasks:
            - name: print-3
              taskSpec: (...)
              runAfter: []
              timeout: 525600m
          iterateParam: param-loop-item
  timeout: 525600m
```
Note the `runAfter: [condition-1]`, while there is clearly no `condition-1` there.

Note that only the loop is affected, not the `print-2` task outside of loop, and that it doesn't matter if `.after` is called on the loop itself or one of its tasks.

## What did you expect to happen:

```yaml
kind: PipelineRun
metadata:
  name: empty-loop
  annotations: (...)
spec:
  params:
  - name: param
    value: '["a", "b", "c"]'
  pipelineSpec:
    params:
    - name: param
      default: '["a", "b", "c"]'
    tasks:
    - name: print-0
      taskSpec: (...)
      timeout: 525600m
    - name: condition-cel
      params:
      - name: outcome
        value: $(tasks.print-0.results.output-value) == "print-0"
      taskRef:
        name: cel_condition
        apiVersion: cel.tekton.dev/v1alpha1
        kind: CEL
      timeout: 525600m
    - name: print-1
      taskSpec: (...)
      when:
      - input: $(tasks.condition-cel.results.outcome)
        operator: in
        values:
        - "true"
      runAfter:
      - print-0
      timeout: 525600m
    - name: print-2
      taskSpec: (...)
      runAfter:
      - print-1
      timeout: 525600m
    - runAfter:
      - print-1
      name: empty-loop-for-loop-2
      params:
      - name: param-loop-item
        value: $(params.param)
      taskSpec:
        apiVersion: custom.tekton.dev/v1alpha1
        kind: PipelineLoop
        spec:
          pipelineSpec:
            params:
            - name: param-loop-item
              type: string
            tasks:
            - name: print-3
              taskSpec: (...)
              runAfter: []
              timeout: 525600m
          iterateParam: param-loop-item
  timeout: 525600m
```

## Additional information:
While the example previously mentioned isn't the minimal reproduce sample, it seems easier to debug the differences on it.

### Analysis
Here are some results I've gathered:

* `dependencies`, that are got from `_get_dependencies`, seem to contain `condition-1` for both the loop AND `print-2`, so it seems that these are rewritten later:
  ```
  {
    'condition-1': {'condition-cel'},
    'condition-cel': {'print-0'},
    'for-loop-2': {'condition-1'},
    'print-2': {'condition-1'}
  }
  ```
* `raw_templates` contain `condition-1` template too
* `_create_pipeline_workflow` returns a yaml without any loop, which contains `print-3` task with `runAfter: [print-1]`
* same after `fix_big_data_passing` and `remove_empty_elements` calls
* the loop appears, already with `runAfter: [condition-1]`, after `_handle_tekton_custom_task` call
  * `custom_task[key]['depends']` returns `[]` for `empty-loop-for-loop-2`
  * at the same place, `custom_task[key]['spec'].get('runAfter', [])` equals to `["condition-1"]`
  * it doesn't seem like that field is changed ever again later on
  * this field seems set back in `_group_to_dag_template`, when constructing raw templates --- the information later used to inline those dependencies is available near the end of that call (without the need to change the signature), it should be moderately easy to use

The conclusion is that `_handle_tekton_custom_task` lacks the same optimization that `_create_pipeline_workflow` has, i.e. removing the conditions which use custom tasks. The same information seems unavailable at that point though.

My suggestion is: "Flatten condition task" should be followed by applying the same flattening too to the spec of `self.loops_pipeline`.

### Minimal sample
```python
from kfp import dsl
from kfp.components import load_component_from_text
from kfp_tekton.tekton import CEL_ConditionOp


class Coder:
  def empty(self):
    return ""


TektonCompiler._get_unique_id_code = Coder.empty


class CelCondition(dsl.Condition):
  def __init__(self, pred: str, name: str = None):
    super().__init__(CEL_ConditionOp(pred).output == 'true', name)


def PrintOp(name: str, msg: str = None):
  if msg is None:
    msg = name
  print_op = load_component_from_text(
  """
  name: %s
  inputs:
  - {name: input_text, type: String, description: 'Represents an input parameter.'}
  outputs:
  - {name: output_value, type: String, description: 'Represents an output paramter.'}
  implementation:
    container:
      image: alpine:3.6
      command:
      - sh
      - -c
      - |
        set -e
        echo $0 > $1
      - {inputValue: input_text}
      - {outputPath: output_value}
  """ % (name)
  )
  return print_op(msg)


@dsl.pipeline("empty-loop")
def condition_1(param: list = ["a", "b", "c"]):
  with CelCondition(f'len({param}) == "print-0"'):
    op1 = PrintOp("print-1")

  loop = dsl.ParallelFor(param)
  with loop:
    op3 = PrintOp("print-3")
    op3.after(op1)


if __name__ == '__main__':
  TektonCompiler().compile(condition_1, __file__.replace('.py', '.yaml'))
```

## Environment:

* SDK Version: 

Tomcli:
Resolves #865 

Argo conditions are wrapped as sub-dag which cannot be dependent with its sub-dag tasks. However, this is not a correct representation for DSL which we have made Tekton to flatten out conditions into non-dag. Therefore, we need to update the dependency logic to depend on the flatten task instead of the removed sub-dag block. 

**Environment tested:**

**Checklist:**


Tomcli:
cc @Udiknedormin

sdk/python/tests/compiler/testdata/loop_with_conditional_dependency.py
```diff
@@ -0,0 +1,69 @@
...
+
+  # ...but breaks for loop or task inside of a loop:
+  loop = dsl.ParallelFor(param)
+  # both A) and B) yield the same result:
+  # A)
+  # loop.after(op1)
+  with loop:
+    op3 = PrintOp("print-3")
+    # B)
+    op3.after(op1)
```
Udiknedormin, line 65: Please include both versions as unit tests.

To me it looks like B) works now but A) still doesn't work. Support for option A) is crucial in order to support empty loops (i.e. loops with no tasks inside).

Udiknedormin (changes requested):

yhwang (approved):
/lgtm

Issue #873, opened 2022-03-09
Metadata Writer Golang package for Tekton Custom Tasks


Tomcli:
/kind feature

**Description:**
When running Tekton Custom tasks, the metadata writer cannot update the MLMD database and push artifacts because Tekton Custom tasks are not new pods. Hence, we need to do the following to enable metadata writer for custom task
1. Update Metadata Writer python server to also watch Tekton `Run`, and update the input/output entry accordingly.
2. Add a Metadata Writer Golang package to push inputs/outputs as artifacts to an object storage using the standard `<bucket> /artifacts/$PIPELINERUN/$PIPELINETASK/<tekton_result_name>.tgz`

**Additional information:**
[Miscellaneous information that will assist in solving the issue.]

Tomcli:
@ScrapCodes maybe you can take a look at number two and implement the new Metadata Writer Golang package? We can discuss in more details tonight. It would be similar to how you pushes log, but this time is for artifacts and need to convert result files into tgz.

Udiknedormin:
@Tomcli Wouldn't that collide with `ARCHIVE_CONFIG_LOG_FILE_NAME`, the default value of which is `step-main.tgz`, which would be the same for a `step-main` result?

PR #876, opened 2022-03-15
feat(objectstore) Metadata object store

ScrapCodes:
Resolves #

Part 2. of #873 

New object store package with readme on how to write artifacts/results of custom tasks to object store.

**Environment tested:**
Tested on IBM Cloud:
ss

**Checklist:**

### DIFF
--- a/tekton-catalog/objectstore/README.md
+++ b/tekton-catalog/objectstore/README.md
@@ -1,0 +1,53 @@
+# objectstore
+
+Connect and store objects in object store. 
+
+**Purpose**: use object store for storing the results of custom tasks.
+
+**For example:**
+
+1. To setup connection.
+```go
+import 	"github.com/kubeflow/kfp-tekton/tekton-catalog/objectstore/pkg/writer"
+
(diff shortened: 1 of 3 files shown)


Tomcli:
Thanks @ScrapCodes. Functionality wise it looks good to me. I'm thinking can the `objectstorelogger` package also be use this new package? In our implementation with Taskrun, both logs and artifacts are pushed to same path. The only exception is that the log file is not compressed. This way we can reduce the number of packages need to import for each controller. What do you think?

It can be a separate PR from this. I will merge this PR for now and open an new issue.

Tomcli:
Thanks @ScrapCodes. Functionality wise it looks good to me. I'm thinking can the `objectstorelogger` package also be use this new package? In our implementation with Taskrun, both logs and artifacts are pushed to same path. The only exception is that the log file is not compressed. This way we can reduce the number of packages need to import for each controller. What do you think?

It can be a separate PR from this. I will merge this PR for now and open an new issue.

PR #911, opened 2022-04-08
feat: Support loop ranges with -ve values.

### LINKED ISSUE #902: Loop.range fails with Cannot determine number of iterations: The from or to parameters was not found in runs
/kind bug

Using the following DSL

``` py
from kfp import dsl, components
from kfp_tekton.tekton import Loop

op1_yaml = '''\
name: 'my-in-coop1'
inputs:
- {name: item, type: Integer}
- {name: param, type: Integer}
implementation:
    container:
        image: library/bash:4.4.23
        command: ['sh', '-c']
        args:
        - |
          set -e
          echo op1 "$0" "$1"
        - {inputValue: item}
        - {inputValue: param}
'''


@dsl.pipeline(name='pipeline')
def pipeline():
    with Loop.range(start=-1, step=1, end=1) as item:
        op1_template = components.load_component_from_text(op1_yaml)
        op1 = op1_template(item, 10)

if __name__ == '__main__':
  from kfp_tekton.compiler import TektonCompiler as Compiler
  from kfp_tekton.compiler.pipeline_utils import TektonPipelineConf
  tekton_pipeline_conf = TektonPipelineConf()
  tekton_pipeline_conf.set_tekton_inline_spec(True)
  # tekton_pipeline_conf.set_resource_in_separate_yaml(True)
  Compiler().compile(pipeline, __file__.replace('.py', '.yaml'), tekton_pipeline_conf=tekton_pipeline_conf)

```

pipeline loop fails in runtime with the error message 

> "type":"Succeeded",
> "status":"False",
> "lastTransitionTime":"2022-03-31T22:22:55Z",
> "reason":"PipelineLoopValidationFailed",
> "message":"Cannot determine number of iterations: The from or to parameters was not found in runs"
> 

This also happens for the following cases
- (start=-1, step=1, end=0)
- (start=-1, step=0, end=0)
- (start=1, step=1, end=-1)
- (start=0, step=1, end=-1)
- (start=-1, step=1, end=-1)
- (start=1, step=0, end=-1)
- (start=0, step=0, end=-1)
- (start=1, step=-1, end=-1)
- (start=0, step=-1, end=-1)

I expect the loop to iterate successfully.

**Additional information:**

**Environment:**

* SDK Version: 

ScrapCodes:
Resolves #902

1.  Support loop ranges with -ve values
2. Added validations and corresponding tests.

**Environment tested:**

**Checklist:**

### DIFF
--- a/tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go
+++ b/tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go
@@ -778,9 +778,9 @@ func (c *Reconciler) updatePipelineRunStatus(ctx context.Context, iterationEleme
 func computeIterations(run *v1alpha1.Run, tls *pipelineloopv1alpha1.PipelineLoopSpec) (int, []interface{}, error) {
 	// Find the iterate parameter.
 	numberOfIterations := -1
-	from := -1
-	step := -1
-	to := -1
+	from := 0
+	step := 1
+	to := 0
 	iterationElements := []interface{}{}
 	iterationParamStr := ""
 	iterationParamStrSeparator := ""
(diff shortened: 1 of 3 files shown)


tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go
```diff
@@ -854,20 +854,25 @@ func computeIterations(run *v1alpha1.Run, tls *pipelineloopv1alpha1.PipelineLoop
 			}
 		}
 	}
-	if from != -1 || to != -1 {
-		if from == -1 || to == -1 {
-			return 0, iterationElements, fmt.Errorf("The from or to parameters was not found in runs")
+	if from != to {
```
yhwang, line 857: do we handle `from == to` somewhere?

ScrapCodes: from == to is not an error at the moment. Do you think it should be?
Actually, if it is 0, it means it is unset.

yhwang: from=0, to=0, step=1 would not be handled by the if condition here, and how will the controller handle it?

ScrapCodes: It is an unset state. i.e. loop is not numeric range.

yhwang: I guess you didn't get my point. ha. the from=0, to=0 could be set in yaml. and how about from=1, to=1, step=1.
my point is those are possible values that you can set from yaml. as long as the controller can handle it well, then it would be good. I just point out the condition here doesn't cover those cases.

ScrapCodes: We don't throw error in that case. What do you expect to happen?

ScrapCodes: The case `from == to`, is handled by controller by not running the loop, I have confirmed this behaviour. This behaviour is unchanged, there is no error reported either. By handling that case my guess was, you are suggesting some error reporting to user.

yhwang: > The case from == to, is handled by controller by not running the loop, I have confirmed this behaviour. This behaviour is unchanged, there is no error reported either.


Thanks, this answers my question. based on the requirement from the issue. no error is expected for this case. If you can add some test cases which address the cases listed in the issue, that would be great.

ScrapCodes:
@yhwang, @Udiknedormin Please take a look!

yhwang (approved):
/lgtm thanks for adding the test cases @ScrapCodes

PR #919, opened 2022-04-16
fix(SDK): Update sanitization logic to sync with kfp dsl

### LINKED ISSUE #914: Tasks and custom tasks have same long name outputs created mismatched parameters
/kind bug

When a pipeline with both tasks and custom tasks referring to the same long name outputs, the pipeline parameters show a mismatched sanitized params for the custom tasks.

**Additional information:**

**Environment:**

* SDK Version: 

Tomcli:
Resolves #914 

For kfp output parameter sanitization, the length is being trimmed by the original kfp dsl sanitization function by length and later on reduced the number of repeated dashes. However, in our Tekton compiler sanitization, it reduced the number of  repeated dashes first then sanitize by the length. This can make certain edge cases where long name output parameters that have multiple repeated dashes can be resulted with a different length in the Tekton results vs Tekton input parameters. Hence, we need to update the sanitization to work the same way as KFP DSL.

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_k8s_helper.py
+++ b/sdk/python/kfp_tekton/compiler/_k8s_helper.py
@@ -54,9 +54,6 @@ def sanitize_k8s_name(name,
     if not allow_slash:
       k8s_name = re.sub('[/]', '-', k8s_name)
 
-    # replace duplicate dashes, strip enclosing dashes
-    k8s_name = re.sub('-+', '-', k8s_name).strip('-')
-
     # truncate if length exceeds max_length
     max_length = max_length - suffix_space
     if len(k8s_name) > max_length:
@@ -64,6 +61,9 @@ def sanitize_k8s_name(name,
         k8s_name = k8s_name[len(k8s_name) - max_length:].strip('-')
       else:
         k8s_name = k8s_name[:max_length].rstrip('-')
+
+    # replace duplicate dashes, strip enclosing dashes
+    k8s_name = re.sub('-+', '-', k8s_name).strip('-')
 
     return k8s_name
 
--- a/sdk/python/kfp_tekton/compiler/_op_to_template.py
+++ b/sdk/python/kfp_tekton/compiler/_op_to_template.py
@@ -364,10 +364,12 @@ def _process_base_ops(op: BaseOp):
     """
 
     # map param's (unsanitized pattern or serialized str pattern) -> input param var str
-    map_to_tmpl_var = {
-        (param.pattern or str(param)): '$(inputs.params.%s)' % param.full_name  # Tekton change
-        for param in op.inputs
-    }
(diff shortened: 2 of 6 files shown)


Tomcli:
cc @Udiknedormin 
This quick fix can address the local compile scenario. I'm not sure will the other env generates the dsl differently. I will try to add few different test cases when I came back next week since I see some inconsistent behaviors with my current set of test cases. 

As of whether to sanitize the `VariableStore`, I could handle it in another PR. `VariableStore` can generate a result name that can be 64 chars, but the SDK will sanitize the input parameters into 57 chars then trim down furthur on any repeated dashes with this PR.

ckadner:
Hi Tommy, 

the order of steps in the sanitation code was chosen that we could use the max allowed number of characters for the sanitized name. If we change to order to remove duplicate dashes after we trim for size, we will get shorter names i.e. we are "losing" some characters. This may not seem important, but for long names that get generated from appending other names, like loop parameters etc the last few characters could become important, i.e. if they are indices like `my-***-loop-param-1`, `my-***-loop-param-2`. I understand the SDK and Backend/Tekton need to create consistent results. Should we consider updating the backend logic instead, to remove duplicate dashes before trimming to size?

Tomcli:
> Hi Tommy,
> 
> the order of steps in the sanitation code was chosen that we could use the max allowed number of characters for the sanitized name. If we change to order to remove duplicate dashes after we trim for size, we will get shorter names i.e. we are "losing" some characters. This may not seem important, but for long names that get generated from appending other names, like loop parameters etc the last few characters could become important, i.e. if they are indices like `my-***-loop-param-1`, `my-***-loop-param-2`. I understand the SDK and Backend/Tekton need to create consistent results. Should we consider updating the backend logic instead, to remove duplicate dashes before trimming to size?

Thanks @ckadner 
The problem happens here is when a regular task is sanitized, they are sanitize with `allow_capital_underscore` flag on first because the loop and dsl arguments still has capital and underscore letters. Once it's passed down to final post processing. The parameter is sanitized again and ended up with fewer than the default 57 characters.

The issue with custom task is that it was sanitized without `allow_capital_underscore` flag at first, if we try to follow the same process and try to sanitize different custom task parameters just like regular task. It will not solve the core problem as some users can explicitly call the `sanitize_k8s_name` function to sanitize their Op name in the DSL code first which will recreate this same issue. Hence I proposed to make the `sanitize_k8s_name` itself to sanitize parameters with consistent character results despite the `allow_capital_underscore` flag is on or not.

I did try to add logics when it sees `-_` and `allow_capital_underscore` is on, we only trim the hyphen character, but that still trigger some inconsistent results when we are using big data passing with loops since big data passing format the intermediate parameters with hyphen where the loop dsl parameters is underscore during the compile phase.

yhwang (approved):
/lgtm

PR #920, opened 2022-04-19
feat(sdk): Add helper class AddOnGroup

### LINKED ISSUE #912: Create new SDK helper functions for embedding exception handlers as part of the DSL
/kind feature

**Description:**
The exception handlers in Watson Studio pipelines are custom tasks that ran inside the exit handle. The exception handlers spec is a Tekton pipeline spec that trigger another pipelinerun (similar to the sub pipeline concept for pipelineloop, but for exit handlers).

Thus, we need to create new SDK helper functions similar to the `loop()` class where this new function can use python `with` block to construct a sub-pipeline spec using python `with` enter and exit logics.

**Additional information:**

yhwang:
add a helper class for users to write a custom OpsGroup which maps to a custom task
in Tekton.

Resolves #912

Add a base class for custom OpsGroup and provide detailed usage under `Advanced User Guide`
**Environment tested:**

* Python Version (use `python --version`): 3.9

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/tekton.py
+++ b/sdk/python/kfp_tekton/tekton.py
@@ -341,11 +341,15 @@ class AddOnGroup(dsl.OpsGroup):
                 api_version: str = DEFAULT_APIVERSION,
                 is_finally: bool = False,
                 parallelism: int = None,
-                params: Dict[str, Union[dsl.PipelineParam, str, int]] = {}):
+                params: Dict[str, Union[dsl.PipelineParam, str, int]] = {},
+                annotation: Dict[str, str] = {},
+                labels: Dict[str, str] = {}):
         self.task_type = task_type  # not been used yet
         self.kind = kind
         self.api_version = api_version
         self.params = params  # for spec.params
+        self.annotation = annotation  # for metadata.annotation
+        self.labels = labels  # for metadata.labels
         if is_finally:
             pl = dsl._pipeline.Pipeline.get_default_pipeline()
             if pl.groups[-1].type != 'pipeline':
--- a/sdk/python/kfp_tekton/tekton.py
+++ b/sdk/python/kfp_tekton/tekton.py
@@ -326,6 +326,16 @@ class AddOnGroup(dsl.OpsGroup):
     DEFAULT_KIND = 'AddOnGroup'
     DEFAULT_APIVERSION = 'custom.tekton.dev/v1alpha1'
 
+    @classmethod
+    def create_internal_param(cls, name: str, value: Optional[str] = None, param_type: Optional[str] = None) -> dsl.PipelineParam:
+        """
+        If a PipelineParam is used for underlying custom task and should not show up in spec.params, use this
+        function to create a PipelineParam and add it into the AddOnGroup.params.
+        """
+        rev = dsl.PipelineParam(name=name, value=value, param_type=param_type)
+        rev.addon_param = True
+        return rev
+
     def __init__(self, task_type: str = RUN_TYPE,
                 kind: str = DEFAULT_KIND,
                 api_version: str = DEFAULT_APIVERSION,
@@ -335,7 +345,7 @@ class AddOnGroup(dsl.OpsGroup):
         self.task_type = task_type  # not been used yet
         self.kind = kind
         self.api_version = api_version
-        self.params = params  # param dict
+        self.params = params  # for spec.params
         if is_finally:
             pl = dsl._pipeline.Pipeline.get_default_pipeline()
             if pl.groups[-1].type != 'pipeline':
@@ -348,6 +358,6 @@ class AddOnGroup(dsl.OpsGroup):
         """provide a post-hook api to update the task YAML"""
         return task_yaml
 
-    def post_param(self, params: List):
+    def post_params(self, params: List):
         """provide a post-hook api to update params"""
         return params
--- a/sdk/python/kfp_tekton/tekton.py
+++ b/sdk/python/kfp_tekton/tekton.py
@@ -311,3 +311,39 @@ class Loop(dsl.ParallelFor):
   def enumerate(self) -> dsl.ParallelFor:
     self.call_enumerate = True
     return self
+
+
+class AddOnGroup(dsl.OpsGroup):
+    """
+    Represents an AddOn group containing ops and group of OpsGroups.
+    This class is the base class for a customized OpsGroup. Users can
+    develop their own OpsGroup. The customized OpsGroup maps to a
+    custom task in Tekton.
+    """
+    TYPE_NAME = 'addon_group'
+    TASK_TYPE = 'task'
+    RUN_TYPE = 'run'  # means custom task
+    DEFAULT_KIND = 'AddOnGroup'
+    DEFAULT_APIVERSION = 'custom.tekton.dev/v1alpha1'
+
+    def __init__(self, task_type: str = RUN_TYPE,
+                kind: str = DEFAULT_KIND,
+                api_version: str = DEFAULT_APIVERSION,
+                is_finally: bool = False,
+                parallelism: int = None,
+                params: List[Union[dsl.PipelineParam, int, str]] = []):
+        self.task_type = task_type  # not been used yet
+        self.kind = kind
+        self.api_version = api_version
+        self.params = params  # list of PipelineParams, or value
+        if is_finally:
+            pl = dsl._pipeline.Pipeline.get_default_pipeline()
+            if pl.groups[-1].type != 'pipeline':
+                raise ValueError(
+                    'You can only create a finally OpsGroup under the root OpsGroup of a Pipeline')
+        self.is_finally = is_finally  # a regular task or finally task
+        super().__init__(self.TYPE_NAME, parallelism=parallelism)
+
+    def post_update(self, task_yaml: dict) -> dict:
+        """provide a post-hook api to update the task YAML"""
+        return task_yaml
--- a/sdk/python/kfp_tekton/tekton.py
+++ b/sdk/python/kfp_tekton/tekton.py
@@ -342,13 +342,13 @@ class AddOnGroup(dsl.OpsGroup):
                 is_finally: bool = False,
                 parallelism: int = None,
                 params: Dict[str, Union[dsl.PipelineParam, str, int]] = {},
-                annotation: Dict[str, str] = {},
+                annotations: Dict[str, str] = {},
                 labels: Dict[str, str] = {}):
         self.task_type = task_type  # not been used yet
         self.kind = kind
         self.api_version = api_version
         self.params = params  # for spec.params
-        self.annotation = annotation  # for metadata.annotation
+        self.annotations = annotations  # for metadata.annotation
         self.labels = labels  # for metadata.labels
         if is_finally:
             pl = dsl._pipeline.Pipeline.get_default_pipeline()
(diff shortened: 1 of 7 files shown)


yhwang:
/cc @Tomcli

@Udiknedormin

yhwang:
@Udiknedormin I added documentation under `Advanced User Guide` [here](https://github.com/kubeflow/kfp-tekton/blob/179c2232c00476b32111c4d91f46ed6b8cf50c1e/guides/advanced_user_guide.md#custom-task-for-opsgroup)
You can try out the `AddOnGroup` class and see if its behaviors and functions fit your needs. By extending that base class, you can use your new class based on your implementation.

Udiknedormin:
Thank you, @yhwang. I will check if it can generate the code we need.

sdk/python/kfp_tekton/tekton.py
```diff
@@ -311,3 +311,39 @@ def __enter__(self) -> Union[Tuple[TektonLoopIterationNumber, LoopArguments], _f
...
+    TASK_TYPE = 'task'
+    RUN_TYPE = 'run'  # means custom task
+    DEFAULT_KIND = 'AddOnGroup'
+    DEFAULT_APIVERSION = 'custom.tekton.dev/v1alpha1'
+
+    def __init__(self, task_type: str = RUN_TYPE,
+                kind: str = DEFAULT_KIND,
+                api_version: str = DEFAULT_APIVERSION,
+                is_finally: bool = False,
+                parallelism: int = None,
```
Udiknedormin, line 333: Doesn't seem to change anything when I set it...

sdk/python/kfp_tekton/tekton.py
```diff
@@ -311,3 +311,39 @@ def __enter__(self) -> Union[Tuple[TektonLoopIterationNumber, LoopArguments], _f
...
+    TYPE_NAME = 'addon_group'
+    TASK_TYPE = 'task'
+    RUN_TYPE = 'run'  # means custom task
+    DEFAULT_KIND = 'AddOnGroup'
+    DEFAULT_APIVERSION = 'custom.tekton.dev/v1alpha1'
+
+    def __init__(self, task_type: str = RUN_TYPE,
+                kind: str = DEFAULT_KIND,
+                api_version: str = DEFAULT_APIVERSION,
+                is_finally: bool = False,
```
Udiknedormin, line 332: I don't think that's how we used to generate `finally` in kfp-tekton for usual tasks. I'm not sure if the DSL users will be ok with that...

I would expect to be able to write sth like:
```python
exit_op = Handler()
with exit_op:
  exit_handler_op_0 = ...
  exit_handler_op_1 = ...

with dsl.ExitHandler(exit_op):
  main_op_0 = ...
  main_op_1 = ...
```
which would produce sth like:
```yaml
tasks:
- name: main-op-0
  ...
- name: main-op-1
  ...
finally:
- name: addon-group-1
  params: ...
  taskSpec:
    apiVersion: custom.tekton.dev/v1alpha1
    kind: Handler
    spec:
      pipelineSpec:
        params: ...
        tasks:
        - name: exc-handler-op-0
          ...
        - name: exc-handler-op-1
          ...
```

...but it doesn't seem to work, failing with `AttributeError: 'Handler' object has no attribute 'dependent_names'`. When I add that fields in the constructor, the following is produced:
```yaml
tasks:
- name: main-op-0
  ...
- name: main-op-1
  ...
- name: addon-group-1
  params: ...
  taskSpec:
    apiVersion: custom.tekton.dev/v1alpha1
    kind: Exception
    spec:
      pipelineSpec:
        params: ...
        tasks:
        - name: exc-handler-op-0
          ...
        - name: exc-handler-op-1
          ...
```

sdk/python/kfp_tekton/tekton.py
```diff
@@ -311,3 +311,39 @@ def __enter__(self) -> Union[Tuple[TektonLoopIterationNumber, LoopArguments], _f
...
+                parallelism: int = None,
+                params: Dict[str, Union[dsl.PipelineParam, str, int]] = {}):
```
Udiknedormin, line 334: I can see that `params` here are the ones that are both for call/instantiation and `spec` --- but in some cases we need to separate the two, just like in range loops: `from`-`step`-`end` for call, but only one param for `spec`.

Initially I thought that maybe the `post_update` can fix that, but I can't see a way to access those spec-parameters inside of the `pipelineSpec`.

I would expect to be able to get sth like:
```python
with Handler(from, step, end) as params:
  item = params.item # or: params['item']
  exit_op_0 = PrintOp(item) 
```

Udiknedormin (changes requested):

sdk/python/kfp_tekton/tekton.py
yhwang: do you mean `parallelism`? for `parallelism`, it's from OpsGroup. I added it to be compatible with OpsGroup. In the tekton compiler, I don't have any special handling for it now.

sdk/python/kfp_tekton/tekton.py
yhwang: When adding `AddOnGroup`, I made it be an `OpsGroup`, which means it can be used to mimic `ExitHandler`. For `ExitHandler`, its constructor tasks a `BaseOp`, but not` OpsGroup`.

Here is an example of using `is_finally=True`:
```python
class Exception(AddOnGroup):
  """Exception Handler"""

  def __init__(self, params: Dict[str, Union[dsl.PipelineParam, str, int]] = {}):
    params['pipelinerun_name']= '$(context.pipelineRun.name)'
    super().__init__(kind='Exception', params=params, api_version='custom.tekton.dev/v1alpha1', is_finally=True)


@dsl.pipeline(
    name='exception-handler',
    description='Addon sample'
)
def download_and_print(url: str = 'gs://ml-pipeline-playground/shakespeare1.txt'):
    """A sample pipeline showing exit handler."""

    echo_op('echo task!')

    with Exception():
        task1()
        task2()
```

and the output YAML should contain the finally section like this:
```yaml
    finally:
    - name: addon-group-1
      params:
      - name: pipelinerun_name
        value: $(context.pipelineRun.name)
      taskSpec:
        apiVersion: custom.tekton.dev/v1alpha1
        kind: Exception
        spec:
          pipelineSpec:
            params:
            - name: pipelinerun_name
              type: string
            tasks:
            - name: task1
              ...
            - name: task2
              ....
```

sdk/python/kfp_tekton/tekton.py
yhwang: I think you can override the `__enter__` when you implement your class that extends `AddOnGroup`. just remember to call `super().__enter__`.

sdk/python/kfp_tekton/tekton.py
Udiknedormin: Hmm... I see. Considering how we always have only one `finally` task anyway, it sounds ok.

sdk/python/kfp_tekton/tekton.py
Udiknedormin: @yhwang When I'm trying to do that, I'm hitting a problem with creating a parameter originating from this OpsGroup:
* if I return it as `dsl.PipelineParam('foo')`, it will get an entry in the call, which I can't remove via `post_update` (as `post_update` can only edit the `spec` part); additionally, if I use it in a downstream task, I'm hitting:
  ```
  Traceback (most recent call last):
    File "(...)/add_on_group__test.py", line 373, in <module>
      TektonCompiler().compile(pipeline_1e436c51_937d_47c8_a440_a4711559e589, __file__.replace('.py', '.yaml'))
    File "(...)/kfp_tekton/compiler/compiler.py", line 1509, in compile
      super().compile(pipeline_func, package_path, type_check, pipeline_conf=pipeline_conf)
    File "(...)/lib/python3.9/site-packages/kfp/compiler/compiler.py", line 1175, in compile
      self._create_and_write_workflow(
    File "(...)/kfp_tekton/compiler/compiler.py", line 1642, in _create_and_write_workflow
      pipeline_loop_crs, workflow = self.prepare_workflow(
    File "(...)/kfp_tekton/compiler/compiler.py", line 1574, in prepare_workflow
    File "(...)/kfp_tekton/compiler/compiler.py", line 1467, in _create_workflow
      workflow = self._create_pipeline_workflow(
    File "(...)/kfp_tekton/compiler/compiler.py", line 1194, in _create_pipeline_workflow
      substitute_param = '$(tasks.%s.results.%s)' % (sanitize_k8s_name(pp.op_name), sanitize_k8s_name(pp.name))
    File "(...)/kfp_tekton/compiler/_k8s_helper.py", line 46, in sanitize_k8s_name
      k8s_name = re.sub('[^-_./0-9A-Za-z]+', '-', name)
    File "(...)/lib/python3.9/re.py", line 210, in sub
      return _compile(pattern, flags).sub(repl, string, count)
  TypeError: expected string or bytes-like object
  ```
* if I return it as `dsl.PipelineParam('foo', op_name=self.name)`, I hit:
  ```
  Traceback (most recent call last):
    File "(...)/add_on_group__test.py", line 363, in <module>
      TektonCompiler().compile(pipeline_1e436c51_937d_47c8_a440_a4711559e589, __file__.replace('.py', '.yaml'))
    File "(...)/kfp_tekton/compiler/compiler.py", line 1509, in compile
      super().compile(pipeline_func, package_path, type_check, pipeline_conf=pipeline_conf)
    File "(...)/lib/python3.9/site-packages/kfp/compiler/compiler.py", line 1175, in compile
      self._create_and_write_workflow(
    File "(...)/kfp_tekton/compiler/compiler.py", line 1642, in _create_and_write_workflow
      pipeline_loop_crs, workflow = self.prepare_workflow(
    File "(...)/kfp_tekton/compiler/compiler.py", line 1574, in prepare_workflow
    File "(...)/kfp_tekton/compiler/compiler.py", line 1467, in _create_workflow
      workflow = self._create_pipeline_workflow(
    File "(...)/kfp_tekton/compiler/compiler.py", line 867, in _create_pipeline_workflow
      raw_templates = self._create_dag_templates(pipeline, op_transformers, params)
    File "(...)/kfp_tekton/compiler/compiler.py", line 578, in _create_dag_templates
      inputs, outputs = self._get_inputs_outputs(
    File "(...)/kfp_tekton/compiler/compiler.py", line 708, in _get_inputs_outputs
      upstream_op = pipeline.ops[param.op_name]
  KeyError: 'addon-group-1'
  ```

sdk/python/kfp_tekton/tekton.py
```diff
@@ -341,11 +341,15 @@ def __init__(self, task_type: str = RUN_TYPE,
                 api_version: str = DEFAULT_APIVERSION,
                 is_finally: bool = False,
                 parallelism: int = None,
-                params: Dict[str, Union[dsl.PipelineParam, str, int]] = {}):
+                params: Dict[str, Union[dsl.PipelineParam, str, int]] = {},
+                annotation: Dict[str, str] = {},
```
Udiknedormin, line 345: It seems like the plural (`annotations`) would fit better here?

Udiknedormin (approved):
Looks like adding labels and annotations work fine now. Thank you!

Just a small thing: the argument (and field) is called `annotation` instead of `annotations`. Otherwise --- I think it looks ok and has all the features necessary.

sdk/python/kfp_tekton/tekton.py
yhwang: nice catch and updated!

Tomcli:
thanks @yhwang and thanks @Udiknedormin for your reviews
/lgtm

Issue #800, opened 2021-12-15
Make pipelineloop to break when seeing a pipelineloop break operation task


Tomcli:
/kind feature

**Description:**
Currently, pipelineloop can break the loop when it sees a task is getting skipped. However, it would make more sense in the DSL if we can have it to also break the loop when it sees a specific task prefix name. This should be an addition to the current break feature instead of replacing it.

**Additional information:**
[Miscellaneous information that will assist in solving the issue.]

Tomcli:
Currently the pipeline loop break only happens when the task is skipped as below. We want to extend this logic to also check whether any task with `pipelineloop-break-operation` prefix in the task name is executed. If so, we should break the loop as below as well. This way, we can have multiple break tasks in the DSL and all of them will have the task prefix `pipelineloop-break-operation`. 
https://github.com/kubeflow/kfp-tekton/blob/c4f52e5f1c360e577187be6c447b5afc1c0505cb/tekton-catalog/pipeline-loops/pkg/reconciler/pipelinelooprun/pipelinelooprun.go#L558-L571

wzhanw:
@Tomcli I think breaking the loop when seeing a pipelineloop break operation task is not precise.
I think we need to check its final status and if the task is skipped we should not break the loop.

PR #936, opened 2022-05-11
wait for pipelineloop-break-operation task final status

wzhanw:
Wait for pipelineloop-break-operation task final status before breaking the loop.

Resolves #
https://github.com/kubeflow/kfp-tekton/issues/800#issuecomment-1123277062

wait until the break operation task is done, then cancel the looprun and break the loop.

**Environment tested:**

**Checklist:**


Tomcli:
@wzhanw Will there be a rare case when the custom task is not able to complete, the pipeline will just not break in this case? @ScrapCodes I remember you were trying to do something like this at some points, do you remember what are the challenges you had?

yhwang:
is it possible to have a test case to verify this?

wzhanw:
Hi @Tomcli,  We have already hit such an issue during the test. but it is not easy to reproduce later. in our case, the break operation task is a custom run that does some calculations and store result in Redis.

Tomcli:
@wzhanw I think your approach make sense. I just also want to get @ScrapCodes 's feedbacks since he was trying the other approach before to force the break task to be completed after breaking the loop.

ScrapCodes:
Since we cannot update directly the status of a custom task not owned by the current controller, this seems fine to me. 
/lgtm

PR #942, opened 2022-05-18
fix(backend): refine copy-artifacts step

yhwang:
to avoid result duplication in the copy-artifacts step, add a step
to move result files to /tekton/home/tep-results.

### DIFF
--- a/backend/src/apiserver/template/tekton_template.go
+++ b/backend/src/apiserver/template/tekton_template.go
@@ -294,9 +294,24 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
 					task.TaskSpec.StepTemplate.VolumeMounts = append(task.TaskSpec.StepTemplate.VolumeMounts, loggingVolumeMounts...)
 				}
 
+				moveStep := false
+				if task.TaskSpec.Results != nil && len(task.TaskSpec.Results) > 0 {
+					// move all results to /tekton/home/tep-results to avoid result duplication in copy-artifacts step
+					// TODO: disable eof strip, since no results under /tekton/results after this step
+					moveResults := workflowapi.Step{Container: corev1.Container{
+						Image:   "busybox",
+						Name:    "move-all-results-to-tekton-home",
+						Command: []string{"sh", "-c"},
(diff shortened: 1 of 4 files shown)


Tomcli:
@Udiknedormin FYI we are changing the `copy-artifacts` step to use steps.[*].container.args and command because Tekton will need another init container if we put the script under steps.[*].container.script. This is part of the performance improvement for when we move to Tekton 0.35+.

Please let us know if this break any of the existing code. Thanks

backend/src/apiserver/template/tekton_template.go
```diff
@@ -322,8 +322,24 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
...
 					t.getEnvVar("STRIP_EOF", strconv.FormatBool(stripEOF)),
 				)
+				container.Command = []string{"sh", "-c"}
+				container.Args = []string{artifactScript}
+
+				step := workflowapi.Step{Container: container}
+
+				if task.TaskSpec.Results != nil && len(task.TaskSpec.Results) > 0 {
+					// move all results to /tekton/home/tep-results to avoid result duplication in copy-artifacts step
+					// TODO: this causes eof strip fails, since no results under /tekton/results after this step
```
Tomcli, line 332: can you add if multiple copy steps exist, don't run the eof command injection? Thanks

Tomcli, line 332: can you add if multiple copy steps exist, don't run the eof command injection? Thanks

yhwang: done

backend/src/apiserver/template/tekton_template.go
```diff
@@ -294,9 +294,24 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
 					task.TaskSpec.StepTemplate.VolumeMounts = append(task.TaskSpec.StepTemplate.VolumeMounts, loggingVolumeMounts...)
 				}
 
+				moveStep := false
+				if task.TaskSpec.Results != nil && len(task.TaskSpec.Results) > 0 {
```
Tomcli, line 298: So I think pretty much all the time we will not have results in the last step. Meaning eof script will not execute all the time?

yhwang: yep, because we need to move out those results file to avoid duplication

PR #967, opened 2022-06-01
feat(sdk): Add labels for artifact outputs when tasks are not referred

### LINKED ISSUE #966: Add flag for kfp-tekton compile to indicate component.yaml outputpath is an artifact
/kind feature

**Description:**
When a component.yaml's output is not referred as a inputpath by another component, the KFP-Tekton compiler only will consider it as a parameter + artifact. We need to have some flag to also able to indicate some component wants this output to be big artifact only. 

**Additional information:**

Tomcli:
Resolves #966 

Right now for component.yaml. If the component's output is not used as an artifact, there's no way to let the output file to be stored as artifacts when needed. Therefore, we want to add a label flag for users to label their outputs as artifacts.

In V2 this feature is not needed since every output is an artifact.

**Environment tested:**

**Checklist:**


sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -489,7 +491,8 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
     # Data passing for the task outputs
     appended_taskrun_name = False
     for task_output in task.get('taskSpec', {}).get('results', []):
-        if (task_name, task_output.get('name')) in outputs_tasks:
+        if (task_name, task_output.get('name')) in outputs_tasks or \
+            task_spec.get('metadata', {}).get('labels', {}).get(ARTIFACT_OUTPUT_LABEL_KEY, '') == ARTIFACT_OUTPUT_LABEL_VALUE:
```
Udiknedormin, line 495: So it looks like the "big data" mode can only be turned-on for the whole component, i.e. all of its outputs?

It doesn't sound like a good idea. It's quite typical for some component to return both small outputs (such as status message) and big outputs (such as standard output of the operation being run).

Udiknedormin (changes requested):

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: For big data outputs, we will copy all the outputs into the object storage and only put back the results that are under 3kb.

Tomcli: If you want to only output a particular output, I can refactor the current big data passing logic. But in that case I will need the task output names as part of the label values.

Udiknedormin: Sure, that will do.

The idea you've described is interesting too, but we'd like to have more control over the mechanism, possibly do what you're describing "manually" in a dedicated controller (may improve both performance and error-handling).

Tomcli: I converted the flag to an annotation that contain a list of output names that will be considered as artifact. You can checkout the test case when `data2` is not specified in the annotation, `data2` will get treated as regular parameter output.

Udiknedormin (approved):

Udiknedormin (commented):
I noticed that when both of the outputs are "big", the following code is generated:
```yaml
image: busybox
name: copy-results-artifacts
script: |
  #!/bin/sh
  set -exo pipefail
  TOTAL_SIZE=0
  ARTIFACT_SIZE=`wc -c $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data0 | awk '{print $1}'`
  TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)
  touch $(results.data0.path)
  if [[ $TOTAL_SIZE -lt 3072 ]]; then
    cp $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data0 $(results.data0.path)
  fi
  ARTIFACT_SIZE=`wc -c $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data1 | awk '{print $1}'`
  TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)
  touch $(results.data1.path)
  if [[ $TOTAL_SIZE -lt 3072 ]]; then
    cp $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data1 $(results.data1.path)
  fi
```

If there are two artifact results, both of which are under 3072, then couldn't both of them fit into the result size limit if the copying simply took place in two separate steps?

Tomcli:
> I noticed that when both of the outputs are "big", the following code is generated:
> 
> ```yaml
> image: busybox
> name: copy-results-artifacts
> script: |
>   #!/bin/sh
>   set -exo pipefail
>   TOTAL_SIZE=0
>   ARTIFACT_SIZE=`wc -c $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data0 | awk '{print $1}'`
>   TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)
>   touch $(results.data0.path)
>   if [[ $TOTAL_SIZE -lt 3072 ]]; then
>     cp $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data0 $(results.data0.path)
>   fi
>   ARTIFACT_SIZE=`wc -c $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data1 | awk '{print $1}'`
>   TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)
>   touch $(results.data1.path)
>   if [[ $TOTAL_SIZE -lt 3072 ]]; then
>     cp $(workspaces.gcs-download.path)/artifacts/$ORIG_PR_NAME/$(context.taskRun.name)/data1 $(results.data1.path)
>   fi
> ```
> 
> If there are two artifact results, both of which are under 3072, then couldn't both of them fit into the result size limit if the copying simply took place in two separate steps?

Here we only can know those two artifacts are big during run time, so we need to define the extra step as part of the pipeline definition. If we want to make it into multiple steps, we will need to ask users to give us an estimated size of the artifact similar to how we use the annotation `tekton-result-sizes`.

yhwang (approved):
/lgtm

PR #974, opened 2022-06-07
support multiple containers in task pod

wzhanw:
Resolves #
The current cache didn't cover the case that the task has more than one step (containers).
This PR will handle the case, and change the outputs as a string object map with contain name as the key, and container output as value.

**Environment tested:**

**Checklist:**


wzhanw:
@Tomcli could you please review this PR, this is a normal case for us.  The current cache assume the task has only one step.

Tomcli:
thanks @wzhanw 
/lgtm
/approve

PR #977, opened 2022-06-15
chore(apiserver): Object Storage connection abstraction

rimolive:
N/A

Removes any reference to minio to make object storage access abstract

**Environment tested:**

* Python Version (use `python --version`): 3.10.4
* Tekton Version (use `tkn version`): 0.33.2
* Kubernetes Version (use `kubectl version`): 1.22.3
* OS (e.g. from `/etc/os-release`): Fedora 35

**Checklist:**

### DIFF
--- a/backend/src/apiserver/client/minio.go
+++ b/backend/src/apiserver/client/minio.go
@@ -57,7 +57,7 @@ func CreateMinioClient(minioServiceHost string, minioServicePort string,
 	return minioClient, nil
 }
 
-func CreateMinioClientOrFatal(minioServiceHost string, minioServicePort string,
+func CreateObjectStoreClientOrFatal(minioServiceHost string, minioServicePort string,
 	accessKey string, secretKey string, secure bool, region string, initConnectionTimeout time.Duration) *minio.Client {
 	var minioClient *minio.Client
 	var err error
--- a/manifests/kustomize/base/installs/generic/pipeline-install-config.yaml
+++ b/manifests/kustomize/base/installs/generic/pipeline-install-config.yaml
@@ -17,6 +17,8 @@ data:
   mlmdDb: metadb
   cacheDb: cachedb
   pipelineDb: mlpipeline
+  objectStoreHost: minio-service
+  objectStorePort: "9000"
   bucketName: mlpipeline
   ## defaultPipelineRoot: Optional. Default pipeline root in v2 compatible mode.
   ## https://www.kubeflow.org/docs/components/pipelines/sdk/v2/v2-compatibility/
--- a/v2/objectstore/object_store.go
+++ b/v2/objectstore/object_store.go
@@ -58,7 +58,7 @@ func OpenBucket(ctx context.Context, k8sClient kubernetes.Interface, namespace s
 		sess, err := session.NewSession(&aws.Config{
 			Credentials:      credentials.NewStaticCredentials(cred.AccessKey, cred.SecretKey, ""),
 			Region:           aws.String("minio"),
-			Endpoint:         aws.String(MinioDefaultEndpoint()),
+			Endpoint:         aws.String(ObjectStoreDefaultEndpoint()),
 			DisableSSL:       aws.Bool(true),
 			S3ForcePathStyle: aws.Bool(true),
 		})
@@ -288,23 +288,23 @@ func downloadFile(ctx context.Context, bucket *blob.Bucket, blobFilePath, localF
 
 // The endpoint uses Kubernetes service DNS name with namespace:
 // https://kubernetes.io/docs/concepts/services-networking/service/#dns
-const defaultMinioEndpointInMultiUserMode = "minio-service.kubeflow:9000"
-const minioArtifactSecretName = "mlpipeline-minio-artifact"
+const defaultObjectStoreEndpointInMultiUserMode = "minio-service.kubeflow:9000"
+const objectStoreArtifactSecretName = "mlpipeline-minio-artifact"
 
-func MinioDefaultEndpoint() string {
+func ObjectStoreDefaultEndpoint() string {
 	// Discover minio-service in the same namespace by env var.
 	// https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables
-	minioHost := os.Getenv("MINIO_SERVICE_SERVICE_HOST")
-	minioPort := os.Getenv("MINIO_SERVICE_SERVICE_PORT")
+	minioHost := os.Getenv("OBJECT_STORE_SERVICE_HOST")
+	minioPort := os.Getenv("OBJECT_STORE_SERVICE_PORT")
 	if minioHost != "" && minioPort != "" {
 		// If there is a minio-service Kubernetes service in the same namespace,
-		// MINIO_SERVICE_SERVICE_HOST and MINIO_SERVICE_SERVICE_PORT env vars should
+		// OBJECT_STORE_SERVICE_HOST and OBJECT_STORE_SERVICE_PORT env vars should
 		// exist by default, so we use it as default.
 		return minioHost + ":" + minioPort
 	}
 	// If the env vars do not exist, we guess that we are running in KFP multi user mode, so default minio service should be `minio-service.kubeflow:9000`.
-	glog.Infof("Cannot detect minio-service in the same namespace, default to %s as MinIO endpoint.", defaultMinioEndpointInMultiUserMode)
-	return defaultMinioEndpointInMultiUserMode
+	glog.Infof("Cannot detect minio-service in the same namespace, default to %s as MinIO endpoint.", defaultObjectStoreEndpointInMultiUserMode)
+	return defaultObjectStoreEndpointInMultiUserMode
 }
 
 type minioCredential struct {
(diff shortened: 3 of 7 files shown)


backend/src/apiserver/template/tekton_template.go
```diff
@@ -331,8 +331,8 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
 					t.getObjectFieldSelector("PIPELINERUN", "metadata.labels['tekton.dev/pipelineRun']"),
 					t.getObjectFieldSelector("PODNAME", "metadata.name"),
 					t.getObjectFieldSelector("NAMESPACE", "metadata.namespace"),
-					t.getSecretKeySelector("AWS_ACCESS_KEY_ID", "mlpipeline-minio-artifact", "accesskey"),
```
Tomcli, line 334: For the S3 access and secret key, we want to load it from the secret instead of putting it as raw EnvVar because we don't want to expose the credential values for anyone who can read the pod spec.

rimolive: That's the point of this change. I'm delegating to the kustomize manifests to inject the env var (which can be from a secret) instead of having a hard-coded secret name.

If you believe this is not a good solution, do you have any suggestions?

Tomcli: The reason we need it to load the environment variable from secret because the Kubeflow multi-user integration will populate the object storage credentials as secret into each namespace. Right now I think the end to end test got stuck because the new pods are not getting the correct AWS_ACCESS_KEY_ID information as raw value environment variable in k8s.

rimolive: How does multi-user integration works? Does kfp-tekton create a Deployment object per namespace?

Tomcli: kfp-tekton is not the one creating the user config. For full Kubeflow deployment, there will be a profile controller that initialize user configuration for each new user namespace such as the minio secret information. Is it possible to keep using the `getSecretKeySelector` function but parameterize the secret name `mlpipeline-minio-artifact` and key name `accesskey` instead?

Tomcli: The user minio secret can be configured in the profile controller client over here.
https://github.com/kubeflow/kfp-tekton/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L366-L379

Tomcli:
@rimolive can you also sign the google CLA? https://cla.developers.google.com/clas
For IBM we have an IBM google group that signed the CLA as an org. Not sure does RedHat have similar way to sign it or not.

backend/src/apiserver/template/tekton_template.go
```diff
@@ -331,8 +331,8 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
...
-					t.getSecretKeySelector("AWS_SECRET_ACCESS_KEY", "mlpipeline-minio-artifact", "secretkey"),
+					t.getEnvVar("AWS_ACCESS_KEY_ID", common.GetStringConfig(common.ObjectStoreAccessKey)),
```
Tomcli, line 334: For the values in the `common` library, we should have some default value so that the default deployment and unit tests can fall back to the default behavior.

backend/src/apiserver/template/tekton_template.go
rimolive: > kfp-tekton is not the one creating the user config. For full Kubeflow deployment, there will be a profile controller that initialize user configuration for each new user namespace such as the minio secret information. Is it possible to keep using the `getSecretKeySelector` function but parameterize the secret name `mlpipeline-minio-artifact` and key name `accesskey` instead?

That was my initial plan, but I did that to be parametrized from the kustomize manifests. I'll setup a multi-user environment to understand how it works, and will use my custom image to see how it behaves.

Tomcli:
> > @rimolive can you also sign the google CLA? https://cla.developers.google.com/clas For IBM we have an IBM google group that signed the CLA as an org. Not sure does RedHat have similar way to sign it or not.
> 
> I already signed the google CLA. Is it something I need to make the PR bot to check again?

I triggered the rescan on my end. The google CLA is looking good. Thanks.

backend/src/apiserver/template/tekton_template.go
Tomcli: sure that sounds good to me

rimolive: Cool, let me know if we can merge this PR and I'll resolve this conversation.

Tomcli: yes, we just need to set back the default values such as `mlpipeline-minio-artifact` for these secret name. So it won't break our current end to end test

backend/src/apiserver/template/tekton_template.go
```diff
@@ -331,8 +331,8 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
 					t.getObjectFieldSelector("PIPELINERUN", "metadata.labels['tekton.dev/pipelineRun']"),
 					t.getObjectFieldSelector("PODNAME", "metadata.name"),
 					t.getObjectFieldSelector("NAMESPACE", "metadata.namespace"),
-					t.getSecretKeySelector("AWS_ACCESS_KEY_ID", "mlpipeline-minio-artifact", "accesskey"),
-					t.getSecretKeySelector("AWS_SECRET_ACCESS_KEY", "mlpipeline-minio-artifact", "secretkey"),
+					t.getEnvVar("AWS_ACCESS_KEY_ID", common.GetStringConfigWithDefault(common.ObjectStoreAccessKey, "mlpipeline-minio-artifact")),
```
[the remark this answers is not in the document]
rimolive: I was trying to look at a simple solution to do that logic, but I'm not yet confident with the result. I'm testing this change in my dev environment to ensure it works, and if the solution is good.

backend/src/apiserver/template/tekton_template.go
```diff
@@ -299,7 +299,7 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
 					// move all results to /tekton/home/tep-results to avoid result duplication in copy-artifacts step
 					// TODO: disable eof strip, since no results under /tekton/results after this step
 					moveResults := workflowapi.Step{
-						Image:   "busybox",
+						Image:   "quay.io/rmartine/busybox",
```
anishasthana, line 302: Is there a shared busybox location in quay that we can refer to instead of one under your personal profile? Barring that, is there a kfp-tekton/opendatahub location we could push to?

rimolive: Actually, this shouldn't be part of the code since this was only to make it work for my environment. I'll revert this change.

backend/src/apiserver/template/tekton_template.go
rimolive: So I sent the update, let me know if that works.

manifests/kustomize/base/installs/generic/pipeline-install-config.yaml
```diff
@@ -17,6 +17,8 @@ data:
   mlmdDb: metadb
   cacheDb: cachedb
   pipelineDb: mlpipeline
+  objectStoreHost: minio-service
+  objectStorePort: 9000
```
Tomcli, line 21:
```suggestion
  objectStorePort: "9000"
```
env variable has to be string, not number.

v2/objectstore/object_store.go
```diff
@@ -288,23 +288,23 @@ func downloadFile(ctx context.Context, bucket *blob.Bucket, blobFilePath, localF
...
+const defaultObjectStoreEndpointInMultiUserMode = "minio-service.kubeflow:9000"
+const objectStoreArtifactSecretName = "mlpipeline-minio-artifact"
 
-func MinioDefaultEndpoint() string {
+func ObjectStoreDefaultEndpoint() string {
 	// Discover minio-service in the same namespace by env var.
 	// https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables
-	minioHost := os.Getenv("MINIO_SERVICE_SERVICE_HOST")
-	minioPort := os.Getenv("MINIO_SERVICE_SERVICE_PORT")
+	minioHost := os.Getenv("OBJECT_STORE_SERVICE_HOST")
```
Tomcli, line 297: Just wondering, are we still want to change the env variable to `OBJECT_STORE_SERVICE_HOST`? If so we need to set the `OBJECT_STORE_SERVICE_HOST` as part of our manifest. The default object store integration for open source has the `minio-service` name.

Tomcli: @rimolive just want to confirm, you want to change the name to `OBJECT_STORE_SERVICE_HOST` for the v2 (next gen) object store only?

backend/src/apiserver/client/minio.go
```diff
@@ -57,7 +57,7 @@ func CreateMinioClient(minioServiceHost string, minioServicePort string,
 	return minioClient, nil
 }
 
-func CreateMinioClientOrFatal(minioServiceHost string, minioServicePort string,
+func CreateObjectStoreClientOrFatal(minioServiceHost string, minioServicePort string,
```
anishasthana, line 60: Is it be design that we still have minioServicePort and stuff in here?

PR #993, opened 2022-07-07
Add alpha feature to output big data passing file path instead of task run name

### LINKED ISSUE #987: Big data passing output result path instead of only the task name
/kind feature

**Description:**
Currently Big data passing output the task name for incoming tasks to construct the result path as its input. We can to change the output itself to a path so that the incoming tasks won't have extra logics to construct paths.

**Additional information:**

Tomcli:
Resolves #987 

Currently Big data passing output the task name for incoming tasks to construct the result path as its input. We can to change the output itself to a path so that the incoming tasks won't have extra logics to construct paths.

Things we found that each task has its own workspace prefix, so the workspace variable is still coded as part of the consuming path.

To enable this feature, run
```
export OUTPUT_BIG_DATA_PATH=true
```

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -450,6 +450,22 @@ def big_data_passing_pipeline(name: str, template: dict, inputs_tasks: set(),
                     "name": task.get('name'),
                     "workspace": pipeline_name
                 })
+        artifact_output_list = task.get('taskSpec', {}).get('metadata', {}).get('annotations', {}).get(
+            ARTIFACT_OUTPUTLIST_ANNOTATION_KEY, '')
+        if artifact_output_list:
+            tmp_list = set()
+            for output in json.loads(artifact_output_list):
+                tmp_list.add(sanitize_k8s_name(output))
+            for task_output in task.get('taskSpec', {}).get('results', []):
+                if task_output.get('name') in tmp_list:
+                    if not task.setdefault('workspaces', []):
+                        task['workspaces'].append({
+                          "name": task.get('name'),
+                            "workspace": pipeline_name
+                        })
+                        pipeline_workspaces.add(pipeline_name)
+                        break
+
     if pipeline_name in pipeline_workspaces:
         # Add workspaces to pipeline
         if not pipeline_spec.setdefault('workspaces', []):
@@ -515,7 +531,7 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
             # For child nodes to know the taskrun output path, it has to pass to results via /tekton/results emptydir
                 if not appended_taskrun_path_step:
                     appended_taskrun_path_step = _get_base_step('output-taskrun-path')
-                appended_taskrun_path_step['script'] += 'echo -n "%s/%s/%s" > $(results.%s-path.path)\n' % \
+                appended_taskrun_path_step['command'] += 'echo -n "%s/%s/%s" > $(results.%s-path.path)\n' % \
                                                     (BIG_DATA_MIDPATH, "$(context.taskRun.name)", task_output.get('name'),
                                                     task_output.get('name'))
                 task['taskSpec']['results'].append({"name": "%s-path" % (task_output.get('name')), "type": "string"})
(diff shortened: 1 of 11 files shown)


Tomcli:
cc @Udiknedormin

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -524,21 +527,35 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
...
+            # For child nodes to know the taskrun output path, it has to pass to results via /tekton/results emptydir
+                if not appended_taskrun_path_step:
+                    appended_taskrun_path_step = _get_base_step('output-taskrun-path')
+                if len(appended_taskrun_path_step['command']) <= 2:
+                    appended_taskrun_path_step['command'].append('')
+                appended_taskrun_path_step['command'][-1] += 'echo -n "%s/%s/%s" > $(results.%s-path.path)\n' % \
+                                                    (BIG_DATA_MIDPATH, "$(context.taskRun.name)", task_output.get('name'),
+                                                    task_output.get('name'))
+                task['taskSpec']['results'].append({"name": "%s-path" % (task_output.get('name')), "type": "string"})
+                task['taskSpec']['steps'].append(appended_taskrun_path_step)
```
Udiknedormin, line 540: Looks like a new `output-taskrun-path` step is added for each subsequent result.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -524,21 +527,35 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
...
-                task['taskSpec']['steps'].append(copy_taskrun_name_step)
-                _append_original_pr_name_env(task)
-                appended_taskrun_name = True
+            if env.get('OUTPUT_BIG_DATA_PATH', 'false').lower() == 'true':
+            # For child nodes to know the taskrun output path, it has to pass to results via /tekton/results emptydir
+                if not appended_taskrun_path_step:
+                    appended_taskrun_path_step = _get_base_step('output-taskrun-path')
+                if len(appended_taskrun_path_step['command']) <= 2:
+                    appended_taskrun_path_step['command'].append('')
+                appended_taskrun_path_step['command'][-1] += 'echo -n "%s/%s/%s" > $(results.%s-path.path)\n' % \
```
Udiknedormin, line 536: I was wondering --- could a prefix (or suffix, or format in general) for paths and values be generalized somehow? My immediate thought would be user-generated outputs and complains about not being able to use `FOO` (standardized into `foo`) and `FOO_PATH` (standardized into `foo-path`) in the same task.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -653,7 +684,8 @@ def append_taskrun_params(task_name_append: str):
     task.get("taskSpec", {})['params'] = [
         param for param in task_spec.get('params', [])
         if (task_name, param.get('name')) not in inputs_tasks or
-            param.get('name').endswith("-trname")
+            param.get('name').endswith("-trname") or
+            param.get('name').endswith("-path")
```
Udiknedormin, line 688: Using suffixes to do some magic in the compiler seems tricky. If the user creates a task with a result named e.g. `asset-path`, it will be interpreted as a path to `asset` result (which does or does not exist), is that correct?

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: thanks, I moved it to the post task output condition

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: These functions are to remove any generated results that are pushed into the pipelineloop or generated by the custom taskspec. I think in the newer Tekton 0.37, we can specify a result type other than `string`. So we can also looking into using result type to specify these special results.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: I parametrize as a python global variable and renamed the suffix to `-datap` (short for data path). If you have a better naming idea I can update it in the global variable.

yhwang (approved):
/lgtm

PR #995, opened 2022-07-11
sdk(feat): Add big data passing with custom volume on kfp-tekton

### LINKED ISSUE #994: Fix big data passing with volume dsl
/kind bug

The big data passing with volume dsl in kfp-tekton didn't get mapped correctly. It should be using pure PVC storage instead of workspace to be align with the argo compiler implementation.

**Additional information:**

**Environment:**

* SDK Version: 

Tomcli:
Resolves #994 

Big data passing with user volume was not properly implemented in kfp-tekton. This feature is specifically for removing argo spec and replace the data passing with k8s pvc mount. Since kfp-tekton already using pvc mount for big data passing, we only need to apply the pipeline config to the workspace config.

Note in the original Argo version, the name of the volume is always fixed like our tekton big data passing since we need consistency names for the pipeline volume. Hence the volume name in kfp-tekton is also fixed to retain the same behavior.
https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/deprecated/compiler/_data_passing_using_volume.py#L54-L55

Also updating the artifact_passing_using_volume.py test cases to the latest kfp version. It needs the `data-volume` pvc be pre-provisioned first with subpath artifact_data/ in order to run the example.

**Environment tested:**

**Checklist:**

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -767,22 +767,22 @@ def _append_original_pr_name_env(task_template):
 
 
 def fix_big_data_passing_using_volume(workflow, pipeline_conf):
-    if workflow['spec'].get('workspaces'):
-        if pipeline_conf.data_passing_method._volume != None:
-            volume_dict = pipeline_conf.data_passing_method._volume.to_dict()
-            if volume_dict.get('persistent_volume_claim'):
-                for workspace in workflow['spec']['workspaces']:
-                    if workspace['name'] == workflow['metadata']['name']:
-                        workspace.pop('volumeClaimTemplate')
-                        temp_vc = {}
-                        for key, value in volume_dict.get('persistent_volume_claim').items():
-                            if value != None:
-                                if key == 'claim_name':
-                                    temp_vc['claimName'] = value
-                                if key == 'read_only':
-                                    temp_vc['readOnly'] = value
-                            workspace['persistentVolumeClaim'] = temp_vc
-                            if pipeline_conf.data_passing_method._path_prefix != None:
-                                workspace['subPath'] = pipeline_conf.data_passing_method._path_prefix
-                            break
+    if workflow['spec'].get('workspaces') and pipeline_conf.data_passing_method._volume is not None:
+        volume_dict = pipeline_conf.data_passing_method._volume.to_dict()
+        volume_dict_pvc = volume_dict.get('persistent_volume_claim')
+        if volume_dict_pvc:
+            for workspace in workflow['spec']['workspaces']:
+                if workspace['name'] == workflow['metadata']['name']:
+                    workspace.pop('volumeClaimTemplate')
+                    temp_vc = {}
+                    for key, value in volume_dict_pvc.items():
+                        if value != None:
+                            if key == 'claim_name':
+                                temp_vc['claimName'] = value
+                            if key == 'read_only':
+                                temp_vc['readOnly'] = value
+                    workspace['persistentVolumeClaim'] = temp_vc
+                    if pipeline_conf.data_passing_method._path_prefix != None:
+                        workspace['subPath'] = pipeline_conf.data_passing_method._path_prefix
+                    break
     return workflow
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -764,3 +764,25 @@ def _append_original_pr_name_env(task_template):
     for step in task_template['taskSpec']['steps']:
         if step['name'] == 'main':
             _append_original_pr_name_env_to_step(step)
+
+
+def fix_big_data_passing_using_volume(workflow, pipeline_conf):
+    if workflow['spec'].get('workspaces'):
+        if pipeline_conf.data_passing_method._volume != None:
+            volume_dict = pipeline_conf.data_passing_method._volume.to_dict()
+            if volume_dict.get('persistent_volume_claim'):
+                for workspace in workflow['spec']['workspaces']:
+                    if workspace['name'] == workflow['metadata']['name']:
+                        workspace.pop('volumeClaimTemplate')
+                        temp_vc = {}
+                        for key, value in volume_dict.get('persistent_volume_claim').items():
+                            if value != None:
+                                if key == 'claim_name':
+                                    temp_vc['claimName'] = value
+                                if key == 'read_only':
+                                    temp_vc['readOnly'] = value
+                            workspace['persistentVolumeClaim'] = temp_vc
+                            if pipeline_conf.data_passing_method._path_prefix != None:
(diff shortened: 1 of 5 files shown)


Tomcli:
cc @Udiknedormin 
/assign @yhwang

yhwang (approved):

sdk/python/kfp_tekton/compiler/compiler.py
```diff
@@ -1519,6 +1519,9 @@ def _create_workflow(self,
 
     workflow = fix_big_data_passing(workflow, self.loops_pipeline, '-'.join(self._group_names[:-1] + [""]))
 
+    if pipeline_conf and pipeline_conf.data_passing_method != None:
```
Udiknedormin, line 1522: *`is not None`

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -764,3 +764,25 @@ def _append_original_pr_name_env(task_template):
     for step in task_template['taskSpec']['steps']:
         if step['name'] == 'main':
             _append_original_pr_name_env_to_step(step)
+
+
+def fix_big_data_passing_using_volume(workflow, pipeline_conf):
+    if workflow['spec'].get('workspaces'):
+        if pipeline_conf.data_passing_method._volume != None:
```
Udiknedormin, line 771: It's quite difficult to read in this nested form. How about borrowing a Go idiom and go with:
```python
if pipeline_conf.data_passing_method._volume is None:
  return workflow
```
etc?

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -764,3 +764,25 @@ def _append_original_pr_name_env(task_template):
...
+                        for key, value in volume_dict.get('persistent_volume_claim').items():
+                            if value != None:
+                                if key == 'claim_name':
+                                    temp_vc['claimName'] = value
+                                if key == 'read_only':
+                                    temp_vc['readOnly'] = value
+                            workspace['persistentVolumeClaim'] = temp_vc
+                            if pipeline_conf.data_passing_method._path_prefix != None:
+                                workspace['subPath'] = pipeline_conf.data_passing_method._path_prefix
+                            break
```
Udiknedormin, line 787: This `break` looks like it's unconditional --- does it mean only the first item should be processed? If so, a loop seems a bit misleading.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -764,3 +764,25 @@ def _append_original_pr_name_env(task_template):
...
+def fix_big_data_passing_using_volume(workflow, pipeline_conf):
+    if workflow['spec'].get('workspaces'):
+        if pipeline_conf.data_passing_method._volume != None:
+            volume_dict = pipeline_conf.data_passing_method._volume.to_dict()
+            if volume_dict.get('persistent_volume_claim'):
+                for workspace in workflow['spec']['workspaces']:
+                    if workspace['name'] == workflow['metadata']['name']:
+                        workspace.pop('volumeClaimTemplate')
+                        temp_vc = {}
+                        for key, value in volume_dict.get('persistent_volume_claim').items():
```
Udiknedormin, line 778: Inside of `if volume_dict.get('persistent_volume_claim')`, further `volume_dict.get('persistent_volume_claim')` instead of `volume_dict['persistent_volume_claim']` seems redundant. In fact, that call's result could be assigned to a variable and that variable could be checked for being `None` instead --- making it more clear that it will be used later after that initial check.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -764,3 +764,25 @@ def _append_original_pr_name_env(task_template):
...
             _append_original_pr_name_env_to_step(step)
+
+
+def fix_big_data_passing_using_volume(workflow, pipeline_conf):
+    if workflow['spec'].get('workspaces'):
+        if pipeline_conf.data_passing_method._volume != None:
+            volume_dict = pipeline_conf.data_passing_method._volume.to_dict()
+            if volume_dict.get('persistent_volume_claim'):
+                for workspace in workflow['spec']['workspaces']:
+                    if workspace['name'] == workflow['metadata']['name']:
```
Udiknedormin, line 775: That `break` pattern would probably be more clear here, as this looks like a "searching loop".

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -764,3 +764,25 @@ def _append_original_pr_name_env(task_template):
...
+                        workspace.pop('volumeClaimTemplate')
+                        temp_vc = {}
+                        for key, value in volume_dict.get('persistent_volume_claim').items():
+                            if value != None:
+                                if key == 'claim_name':
+                                    temp_vc['claimName'] = value
+                                if key == 'read_only':
+                                    temp_vc['readOnly'] = value
+                            workspace['persistentVolumeClaim'] = temp_vc
```
Udiknedormin, line 784: Mapping lookup + if seems more clear here.

`temp_vc` seems unneeded, as considering the loop only iterates once, it looks like it's never accumulating anything but just holding the single information above (either `claimName` or `readOnly`), so it could even be inlined as a dict literal.

sdk/python/tests/compiler/testdata/artifact_passing_using_volume.py
```diff
@@ -16,35 +16,73 @@
...
+    processor_task = processor_op(producer_task.outputs['output_1'],
+                                  producer_task.outputs['output_2'])
+    consumer_task = consumer_op(processor_task.outputs['output_1'],
+                                processor_task.outputs['output_2'])
 
+    markdown_task = create_component_from_func(func=metadata_and_metrics)()
     # This line is only needed for compiling using dsl-compile to work
-    kfp.dsl.get_pipeline_conf().data_passing_method = volume_based_data_passing_method
+    kfp.dsl.get_pipeline_conf(
+    ).data_passing_method = volume_based_data_passing_method
```
Udiknedormin, line 75: I don't really understand this formatting...

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: In the test case there's only one key, but v1volume can have both claim_name and read_only (so can iterate up to two times and can be more with future changes in v1volume class).

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: Sorry there were some issues when I reformatted this code using vscode, I fixed the indentation now.

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
Tomcli: done

yhwang (approved):
/lgtm

PR #997, opened 2022-07-12
fix(sdk): Support passing directories in step-copy-results-artifacts and step-copy-artifacts. Fixes #996

### LINKED ISSUE #996: Directories passed as InputPath/OutputPath are not uploaded to minio and the pipeline remains hanging in the UI
/kind bug

I developed the following simple pipeline passing a directory between two components as InputPath/OutputPath. The resulting artefact is not correctly uploaded to minio and the pipeline remains hanging in the UI.

```
import kfp
from kfp import dsl
from kfp.components import func_to_container_op
from kfp.components import InputPath, OutputPath
import os

# environment variables for storage
os.environ["DEFAULT_ACCESSMODES"] = "ReadWriteOnce"
os.environ["DEFAULT_STORAGE_SIZE"] = "2Gi"
# os.environ["DEFAULT_STORAGE_CLASS"] = "standard"


def populate_folder(output_folder_path: OutputPath("directory")):
  """
  Populate a directory with two files
  Parameters:
  -----------
  output_folder_path: OutputPath("directory")
    Path to the directory storing the two files
  """
  # import modules
  import os
  # create folder
  os.makedirs(output_folder_path, exist_ok=True)
  # write files in folder
  with open(os.path.join(output_folder_path, "file1.txt"), "w") as stream:
    stream.write("File 1 - WRITE component")
  with open(os.path.join(output_folder_path, "file2.txt"), "w") as stream:
    stream.write("File 2 - WRITE component")


def read_files(input_folder_path: InputPath("directory")):
  """
  Read and print the content of files in a given directory
  Parameters:
  -----------
  input_folder_path: InputPath("directory")
    Path to the directory storing the initial files
  """
  # import modules
  import os
 # loop over files
  for filename in os.listdir(input_folder_path):
    complete_filename = os.path.join(input_folder_path, filename)
    # read and print file
    with open(complete_filename, "r") as stream:
      for line in stream:
        print(line + " + READ component")


# define operators
populate_folder_component = func_to_container_op(
  func=populate_folder
)

read_files_component = func_to_container_op(
  func=read_files
)


# define pipeline
@dsl.pipeline(
  name="FolderPassing"
)
def folder_passing_pipeline():
  populate_folder_task = populate_folder_component()
  read_files_task = read_files_component(populate_folder_task.output)


if __name__ == "__main__":
  TektonCompiler().compile(folder_passing_pipeline, "folder_passing_pipeline.yaml")
```

**Additional information:**
The issues are due to:
- (for the hanging pipeline) the fact that the step-copy-results-artifacts script copies the artifact into a file which is then tentatively read as a string (same issue highlighted in #984 for binary files).
- (for the missing upload) the fact that the step-copy-artifacts checks the existence of the artifact itself by ```if [ -f "$2" ]```, so that directories are neglected.

**Environment:**

* Python Version (use `python --version`): Python 3.10.5
* SDK Version: 1.2.1
* Kubernetes Version (use `kubectl version`): client v1.23.6, server v1.24.0
* OS (e.g. from `/etc/os-release`): macOS Monterey v12.4

FFede0:
Resolves #996 

1. To add support to directories in step-copy-artifacts script, I added check [ -d "$2" ] to allow directories to be transmitted as artifacts

2. To add support to directories in step-copy-results-artifacts script, I check if the provided result (src) is a directory; if so, create a tar.gz archive and compute the ARTIFACT_SIZE on {src}.tar.gz.
Moreover, print on {dst} file the list of files in {src}.tar.gz (namely, the result of `tar -tzf {src}.tar.gz`) instead of copying directly {src} into {dst}

**Environment tested:**

* Python Version (use `python --version`): Python 3.10.5
* Tekton Version (use `tkn version`): kfp-tekton v1.2.1
* Kubernetes Version (use `kubectl version`): client v1.23.6, server v1.24.0
* OS (e.g. from `/etc/os-release`): macOS Monterey v12.4

### DIFF
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -637,21 +637,17 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
                             script += (
                                     'if [ -d ' + src + ' ]; then\n' + 
                                     '  tar -czvf ' + src + '.tar.gz ' + src + '\n' +
-                                    '  ARTIFACT_SIZE=`wc -c %s.tar.gz | awk \'{print $1}\'`\n' % src +
-                                    '  TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
-                                    '  touch ' + dst + '\n' +  # create an empty file by default.
-                                    '  if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
-                                    '    tar -tzf ' + src + '.tar.gz > ' + dst + '\n' +
-                                    '  fi\n'
-                                    'else\n' +
-                                    '  ARTIFACT_SIZE=`wc -c %s | awk \'{print $1}\'`\n' % src +
-                                    '  TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
-                                    '  touch ' + dst + '\n' +  # create an empty file by default.
-                                    '  if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
-                                    '    if ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
-                                    '      cp ' + src + ' ' + dst + '\n' +
-                                    '    fi\n'
-                                    '  fi\n'
+                                    '  SUFFIX=".tar.gz"\n' +
+                                    'fi\n' +
+                                    'ARTIFACT_SIZE=`wc -c %s${SUFFIX} | awk \'{print $1}\'`\n' % src +
+                                    'TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
+                                    'touch ' + dst + '\n' +  # create an empty file by default.
+                                    'if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
+                                    '  if [ -d ' + src + ' ]; then\n' + 
+                                    '    tar -tzf ' + src + ' > ' + dst + '\n' +
+                                    '  elif ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
+                                    '    cp ' + src + ' ' + dst + '\n' +
+                                    '  fi\n' +
                                     'fi\n'
                             )
             copy_results_artifact_step['command'].append(script)
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -634,7 +634,7 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
                         dst = '$(results.%s.path)' % sanitize_k8s_name(result['name'])
                         if artifact_name == result['name'] and src != dst:
                             add_copy_results_artifacts_step = True
-                            copy_results_artifact_step['script'] += (
+                            script += (
                                     'if [ -d ' + src + ' ]; then\n' + 
                                     '  tar -czvf ' + src + '.tar.gz ' + src + '\n' +
                                     '  ARTIFACT_SIZE=`wc -c %s.tar.gz | awk \'{print $1}\'`\n' % src +
--- a/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
+++ b/sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
@@ -635,7 +635,7 @@ def big_data_passing_tasks(prname: str, task: dict, pipelinerun_template: dict,
                         if artifact_name == result['name'] and src != dst:
                             add_copy_results_artifacts_step = True
                             script += (
-                                    'if [ -d ' + src + ' ]; then\n' + 
+                                    'if [ -d ' + src + ' ]; then\n' +
                                     '  tar -czvf ' + src + '.tar.gz ' + src + '\n' +
                                     '  SUFFIX=".tar.gz"\n' +
                                     'fi\n' +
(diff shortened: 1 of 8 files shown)


sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -634,13 +634,23 @@ def append_taskrun_params(task_name_append: str):
...
                         if artifact_name == result['name'] and src != dst:
                             add_copy_results_artifacts_step = True
-                            script += (
-                                    'ARTIFACT_SIZE=`wc -c %s | awk \'{print $1}\'`\n' % src +
-                                    'TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
-                                    'touch ' + dst + '\n' +  # create an empty file by default.
-                                    'if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
-                                    '  if ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
-                                    '    cp ' + src + ' ' + dst + '\n' +
+                            copy_results_artifact_step['script'] += (
```
Tomcli, line 637: We want to move away from tekton script to tekton container spec so that it can create the task pod with fewer init-containers.
```suggestion
                            script += (
```

Tomcli:
@FFede0 do you think there's a way you can reduce the string or determine the output is a directory during pipeline compiling? Our goal is to reduce the amount of scripts as much as possible so it won't hit the Kubernetes etcd size limit of 1MB.

FFede0:
@Tomcli, how much should the script be reduced? Would something like this be more reasonable?
```
'if [ -d ' + src + ' ]; then\n' + 
'  tar -czvf ' + src + '.tar.gz ' + src + '\n' +
'  SUFFIX=".tar.gz"\n' +
'fi\n'
'ARTIFACT_SIZE=`wc -c %s${SUFFIX} | awk \'{print $1}\'`\n' % src +
'TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
'touch ' + dst + '\n' +  # create an empty file by default.
'if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
'  if [ -d ' + src + ' ]; then\n' + 
'    tar -tzf ' + src + ' > ' + dst + '\n' +
'  elif ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
'    cp ' + src + ' ' + dst + '\n' +
'  fi\n' +
'fi\n'
```
I don't know how determining if the output is a directory during pipeline compilation would be possible...as an alternative, I probably see adding a flag to be passed to the compiler. However, I am not sure this is a solution though: it may happen that who writes the pipeline component does not actually know if the data he is receiving are files or directories (e.g., if he is downloading some data from a website).
Please, let me know if this makes sense to you.

Tomcli:
> @Tomcli, how much should the script be reduced? Would something like this be more reasonable?
> 
> ```
> 'if [ -d ' + src + ' ]; then\n' + 
> '  tar -czvf ' + src + '.tar.gz ' + src + '\n' +
> '  SUFFIX=".tar.gz"\n' +
> 'fi\n' +
> 'ARTIFACT_SIZE=`wc -c %s${SUFFIX} | awk \'{print $1}\'`\n' % src +
> 'TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
> 'touch ' + dst + '\n' +  # create an empty file by default.
> 'if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
> '  if [ -d ' + src + ' ]; then\n' + 
> '    tar -tzf ' + src + ' > ' + dst + '\n' +
> '  elif ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
> '    cp ' + src + ' ' + dst + '\n' +
> '  fi\n' +
> 'fi\n'
> ```
> 
> I don't know how determining if the output is a directory during pipeline compilation would be possible...as an alternative, I probably see adding a flag to be passed to the compiler. However, I am not sure this is a solution: it may happen that who writes the pipeline component does not actually know if the data it is receiving are files or directories (e.g., if it is downloading them from a website). Please, let me know if this makes sense to you.

Yeah I think the script you have above is reasonable, so we are less duplicated code in the condition. 
Can you update that code in your PR and apply the changes in my suggestions so the tests can be passed? thanks

FFede0:
@Tomcli, I pushed the script with reduced size as required

Tomcli:
@FFede0 can you run `make unit_test GENERATE_GOLDEN_YAML=True` to regenerate the unit test results?

FFede0:
@Tomcli, done

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
```diff
@@ -635,13 +635,19 @@ def append_taskrun_params(task_name_append: str):
...
+                                    '  tar -czvf ' + src + '.tar.gz ' + src + '\n' +
+                                    '  SUFFIX=".tar.gz"\n' +
+                                    'fi\n' +
+                                    'ARTIFACT_SIZE=`wc -c %s${SUFFIX} | awk \'{print $1}\'`\n' % src +
                                     'TOTAL_SIZE=$( expr $TOTAL_SIZE + $ARTIFACT_SIZE)\n' +
                                     'touch ' + dst + '\n' +  # create an empty file by default.
                                     'if [[ $TOTAL_SIZE -lt 3072 ]]; then\n' +
-                                    '  if ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
+                                    '  if [ -d ' + src + ' ]; then\n' + 
+                                    '    tar -tzf ' + src + ' > ' + dst + '\n' +
```
Tomcli, line 647: It looks like your generated tests doesn't match with the one generated by the test build. Also, I don't think Tekton can pass a directory as the Tekton results, are you sure this command will work in this case?

Tomcli (commented):
BTW, you need to run `make lint` to fix the python lint test

sdk/python/kfp_tekton/compiler/_data_passing_rewriter.py
FFede0: Should I do anything except running ```make unit_test GENERATE_GOLDEN_YAML=True``` to generate the tests? I do not get why the code differs from what I wrote in _data_passing_rewriter.py file

Also, concerning passing directories, the fact that Tekton cannot pass a directory as the Tekton results is precisely the reason why I added the check: writing
```
'  if [ -d ' + src + ' ]; then\n' + 
'    tar -tzf ' + src + '.tar.gz > ' + dst + '\n' +
'  elif ! awk "/[^[:print:]]/{f=1} END{exit !f}" %s; then\n' % src +
'    cp ' + src + ' ' + dst + '\n' +
```
I make sure that, if the result is a directory, it is not copied in ```dst```, but instead only the content  of the compress archive is written in the file (I could have written ```'ls ' + src + ' > ' + dst '\n'``` probably).

You were right, though, there was a mistake that now I fixed

FFede0:
@Tomcli I run the ```make lint``` and fixed two small errors that came from it and from your comment above. Please, let me know what you meant by "your generated tests doesn't match with the one generated by the test build": I opened some yaml files generated by running ```make unit_test GENERATE_GOLDEN_YAML=True``` and it seems to me that the copy-results-artifacts script matches what I wrote in _data_passing_rewriter.py. Am I missing something?

Tomcli:
Thanks @FFede0 for your contribution. This PR looks good. I think we can further reduce the original script so I opened a follow up issue to further optimize this feature.

/lgtm
/approve

PR #1051, opened 2022-09-30
test: Add a test case for the validation

yhwang:

Add a test to verify the performance of webhook
validation. huge amount of edges in DAG causes
a performance issue in tekton earlier. adding
this test case to avoid the performance degradation.

**Checklist:**

### DIFF
--- a/scripts/deploy/iks/test/many-edges.py
+++ b/scripts/deploy/iks/test/many-edges.py
@@ -1,0 +1,51 @@
+# Copyright 2020 kubeflow.org
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


yhwang:
I manually verified this on tekton 0.38.3 and the test failed. The error is the webhook timeout. I also kicked off a run in the toolchain, and it passed.

Webhook timeout is 10s. I set a 5-second check in the script. if kicking off the run gets longer than 5 seconds, the test case fails.

scripts/deploy/iks/test/many-edges.py
```diff
@@ -0,0 +1,51 @@
...
+PRINT_STR = """
+name: print
+description: print a message
+inputs:
+  - {name: msg, type: String}
+implementation:
+  container:
+    image: alpine:3.6
+    command:
+    - echo
```
Tomcli, line 27: i don't think you can run with 'echo' as starting command, maybe something like `sh -c echo ....`?

yhwang: yes, it works. it took about 5 mins to complete the pipelinerun in my kind cluter :-D

And inside the script, it only verify if it can submit the run. it doesn't wait for the run to finish

Tomcli: it looks good to me now

PR #1065, opened 2022-10-26
feat(backend): add new annotation flags to enable artifact tracking at pipeline and task level

### LINKED ISSUE #940: Fine-grained control enabling situational values for global settings
/kind feature

**Description:**
Artifacts tracking can be very useful, but turning it on globally in case it's not required or desired for every single run is wasteful.

Another example for a setting that would benefit fine grained control is stripping end of line characters.
There are pipelines that require the setting to be off to succeed and therefore are not sharable per se with other deployments

The global setting should be the default value, but it would be nice to have additional handles to provide situational values.
Most important to me personally would still be settings on the deployment level. I'm not sure about the need or ramifications for changing the values with each run.

Maybe with a precedence like the following:
global < profile < experiment < pipeline definition < pipeline run

Tomcli:
Resolves #940 

Add new annotation flags to enable artifact tracking at pipeline and task level with backward compatibility.

**Environment tested:**

**Checklist:**


yhwang (approved):
/lgtm

PR #1259, opened 2023-06-22
feat(backend): Source ObjStore Creds from Env in Tekton Template

gmfrasca:
Resolves #

There is a hard-coded reference to a Secret named `mlpipeline-minio-artifact` that contains credentials metadata to the Object Store.  This is the only instance of this, so sourcing the metadata from the apiserver's Env Vars (where it is already provided) will make deployments more flexible and is better practice, in general.

Also updates a viper config variable to match/target what is used across the rest of the project and manifests.  It was also the only instance of this (`_SECRETKEY` rather than `_SECRETACCESSKEY`), so likely was a typo.

**Environment tested:**

* Python Version (use `python --version`): 3.11.1
* Tekton Version (use `tkn version`): v0.31.4
* Kubernetes Version (use `kubectl version`): v1.22
* OS (e.g. from `/etc/os-release`): Fedora 37

**Checklist:**

### DIFF
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -48,8 +48,11 @@ const (
 	TerminateStatus                         string = "TERMINATE_STATUS"
 	MoveResultsImage                        string = "MOVERESULTS_IMAGE"
 	Path4InternalResults                    string = "PATH_FOR_INTERNAL_RESULTS"
+	ObjectStoreCredentialsSecret            string = "OBJECTSTORECONFIG_CREDENTIALSSECRET"
+	ObjectStoreCredentialsAccessKeyKey      string = "OBJECTSTORECONFIG_CREDENTIALSACCESSKEYKEY"
+	ObjectStoreCredentialsSecretKeyKey      string = "OBJECTSTORECONFIG_CREDENTIALSSECRETKEYKEY"
 	ObjectStoreAccessKey                    string = "OBJECTSTORECONFIG_ACCESSKEY"
-	ObjectStoreSecretKey                    string = "OBJECTSTORECONFIG_SECRETKEY"
+	ObjectStoreSecretKey                    string = "OBJECTSTORECONFIG_SECRETACCESSKEY"
 )
 
 func IsPipelineVersionUpdatedByDefault() bool {
@@ -144,6 +147,26 @@ func GetPodNamespace() string {
 
 func GetArtifactImage() string {
 	return GetStringConfigWithDefault(ArtifactImage, DefaultArtifactImage)
+}
+
+func GetObjectStoreAccessKey() string {
+	return GetStringConfig(ObjectStoreAccessKey)
+}
+
+func GetObjectStoreSecretKey() string {
+	return GetStringConfig(ObjectStoreSecretKey)
+}
+
+func GetObjectStoreCredentialsSecretName() string {
+	return GetStringConfigWithDefault(ObjectStoreCredentialsSecret, DefaultObjectStoreCredentialsSecret)
+}
+
+func GetObjectStoreCredentialsAccessKeyKey() string {
+	return GetStringConfigWithDefault(ObjectStoreCredentialsAccessKeyKey, DefaultObjectStoreCredentialsAccessKeyKey)
+}
+
+func GetObjectStoreCredentialsSecretKeyKey() string {
+	return GetStringConfigWithDefault(ObjectStoreCredentialsSecretKeyKey, DefaultObjectStoreCredentialsSecretKeyKey)
 }
 
 func GetMoveResultsImage() string {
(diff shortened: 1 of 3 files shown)


backend/src/apiserver/common/config.go
```diff
@@ -146,6 +146,14 @@ func GetArtifactImage() string {
 	return GetStringConfigWithDefault(ArtifactImage, DefaultArtifactImage)
 }
 
+func GetObjectStoreAccessKey() string {
+	return GetStringConfig(ObjectStoreAccessKey)
```
Tomcli, line 150: By getting the S3 credential from viper, the whole cluster has to share the same S3 credential for all users. Whereas by getting the S3 credential from the secret, because users has full control of their namespaces so they can update the `mlpipeline-minio-artifact` with their own S3 credential without impacting other users.

backend/src/apiserver/template/tekton_template.go
```diff
@@ -354,8 +356,8 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
 					t.getObjectFieldSelector("PIPELINERUN", "metadata.labels['tekton.dev/pipelineRun']"),
 					t.getObjectFieldSelector("PODNAME", "metadata.name"),
 					t.getObjectFieldSelector("NAMESPACE", "metadata.namespace"),
-					t.getSecretKeySelector("AWS_ACCESS_KEY_ID", "mlpipeline-minio-artifact", "accesskey"),
```
Tomcli, line 357: maybe instead of hard code the `mlpipeline-minio-artifact` secret name, we can parameterize it so users can point to other secret name for s3 credentials.

backend/src/apiserver/common/config.go
gmfrasca: (this and the following review feedback items are related so centralizing the discussion here)

This makes sense, and it would be good to parameterize these somehow.  As you mentioned below, the easiest path is probably to create some new env vars for the s3 Secret name, accesskey Key, and secretkey Key.  It is a little clunky, but does remove the hardcoded secret name.  However, this would mean every namespace would need to use identical secret names across Namespaces (ie namespace1 and namespace2 can each have a `foo-s3-secret` with different contents, but the secret *must* always be named `foo-s3-secret` and use the same `accesskey` and `secretkey` keys.

I think that's a step in the right direction but it would be better if a namespace can individually reference an arbitrary secret name in a vacuum (namespace1 references `foo-secret` but namespace2 references `bar-secret`).  I'm not sure of a clean way to incorporate that though, you wouldn't happen to know of a better vector for retrieving these parameters than apiserver env vars, would you?  If not I can push a changeset that implements the parameterized secret name as outline above.

backend/src/apiserver/template/tekton_template.go
```diff
@@ -354,8 +356,8 @@ func (t *Tekton) injectArchivalStep(workflow util.Workflow, artifactItemsJSON ma
...
-					t.getSecretKeySelector("AWS_SECRET_ACCESS_KEY", "mlpipeline-minio-artifact", "secretkey"),
+					t.getEnvVar("AWS_ACCESS_KEY_ID", objectStoreAccessKey),
```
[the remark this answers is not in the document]
gmfrasca: yikes, didn't realize this was getting templated into the pod spec directly, sorry about that.  Upcoming changeset will revert back to `getSecretKeySelector`

Tomcli:
Thanks @gmfrasca, could you rebase this PR with the latest commit so we can test it against the new Tekton v1 API?

backend/src/apiserver/common/config.go
```diff
@@ -48,8 +48,11 @@ const (
 	TerminateStatus                         string = "TERMINATE_STATUS"
 	MoveResultsImage                        string = "MOVERESULTS_IMAGE"
 	Path4InternalResults                    string = "PATH_FOR_INTERNAL_RESULTS"
+	ObjectStoreCredentialsSecret            string = "OBJECTSTORECONFIG_CREDENTIALSSECRET"
+	ObjectStoreCredentialsAccessKeyKey      string = "OBJECTSTORECONFIG_CREDENTIALSACCESSKEYKEY"
+	ObjectStoreCredentialsSecretKeyKey      string = "OBJECTSTORECONFIG_CREDENTIALSSECRETKEYKEY"
 	ObjectStoreAccessKey                    string = "OBJECTSTORECONFIG_ACCESSKEY"
-	ObjectStoreSecretKey                    string = "OBJECTSTORECONFIG_SECRETKEY"
+	ObjectStoreSecretKey                    string = "OBJECTSTORECONFIG_SECRETACCESSKEY"
```
Tomcli, line 55: @rimolive I remember you introduced this new variable. In KFP-Tekton I don't think we are using this variable in open source. Are you okay with this change?

gmfrasca: FWIW I believe this was just a typo that I've corrected here, vanilla KFP has instances of what I've updated it to (`OBJECTSTORECONFIG_SECRETACCESSKEY`) in their manifests, but as @Tomcli noted it doesn't appear to be used anywhere here in KFP-Tekton so it never broke anything.

Tomcli:
/lgtm
/approve
thanks @gmfrasca

PR #1287, opened 2023-07-14
fix(v2:backend): fix importer

yhwang:
the input value of `importer_spec` is incorrect. It was component spec. Changed it to importer spec.

**Checklist:**


backend/src/v2/compiler/tektoncompiler/testdata/importer.yaml
```diff
@@ -83,10 +83,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.uid
-          envFrom:
-          - configMapRef:
-              name: metadata-grpc-configmap
```
Tomcli, line 88: Was this a new standard from kfp v2 or it's just used in this test case? I remember v1 was using env variable for looking up the metadata host.

yhwang: this part is not new. that configmap has been there for 3 yrs.

Tomcli: I see, this was just under the integration test, but not the official deployment.

yhwang:
thanks, later on, I will build a new image for the api-server and update the manifest to point to that image

PR #1313, opened 2023-08-03
v2: add multi-user manifests

yhwang:
Add multi-user manifests and remove cache and metadata-writer.

**Checklist:**

### DIFF
--- a/manifests/kustomize/env/platform-agnostic-tekton/kustomization.yaml
+++ b/manifests/kustomize/env/platform-agnostic-tekton/kustomization.yaml
@@ -8,7 +8,7 @@ resources:
 - ../../third-party/tekton-custom-task
 - ../../third-party/minio/base
 - ../../third-party/mysql/base
-- ../../third-party/kfp-csi-s3
+# - ../../third-party/kfp-csi-s3
 
 # Identifier for application manager to apply ownerReference.
 # The ownerReference ensures the resources get garbage collected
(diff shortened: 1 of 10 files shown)


manifests/kustomize/env/platform-agnostic-tekton/kustomization.yaml
```diff
@@ -8,6 +8,7 @@ resources:
 - ../../third-party/tekton-custom-task
 - ../../third-party/minio/base
 - ../../third-party/mysql/base
+- ../../third-party/kfp-csi-s3
```
Tomcli, line 11: Do we need kfp-csi-s3 for v2? This was mainly meant for copying big artifacts into s3. Do we have any pvc implementation that uses this csi driver by default in v2?

yhwang: no built-in usage for that. but you can create/user volume in your pipeline by using kfp-kubernetes sdk

yhwang: I guess we can put the manifest in, but not have it in the kustomization.yaml

Tomcli: We shouldn't include it by default then, in iks the cos driver has to install separately. If nothing is using it by default we should make users to opt-in if they want to install kfp-csi-s3.

Tomcli: One of the main reason is that cos driver only support iam auth, we can't use default minio auth on cos driver.

PR #1395, opened 2023-10-20
feat(backend): Add CA injection to step-copy-artifacts step. Fixes #1394.

### LINKED ISSUE #1394: Support artifact passing in airgapped environments with self-signed certs
/kind feature

**Description:**
When running kfp-tekton withing a network with it's own pki the api server will fail to connect to the minio with the following error: 

```
x509: certificate signed by unknown authority
```
This is easily over come by mounting the trusted bundle onto the api server pod. However it is not so trivial to do the same for the pods created via tekton in the step-copy-artifact step: 

image

API server would need to have the option to allow for injecting trusted bundles from a configmap into the step-copy-artifact step as that's when the job will write to the object store.

Additional information:

This is mostly from v1 perspective, I haven't confirmed the behavior for v2 (is there a tekton step that engages with s3 similar to v1?)

HumairAK:
Resolves https://github.com/kubeflow/kfp-tekton/issues/1394

This PR adds the option to pull in a trusted bundle that gets mounted to the step-copy artifact steps added by the api server when generating the tekton template. It assumes default location of /etc/ssl/certs. 

**Environment tested:**

* Python Version (use `python --version`): Python 3.11, tested with kfp sdk 1.8
* Tekton Version (use `tkn version`): 0.50
* Kubernetes Version (use `kubectl version`): OC stable-4.13 / k8s 1.26

**Checklist:**

### DIFF
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -43,6 +43,9 @@ const (
 	ArtifactScript                          string = "ARTIFACT_SCRIPT"
 	ArtifactImage                           string = "ARTIFACT_IMAGE"
 	ArtifactCopyStepTemplate                string = "ARTIFACT_COPY_STEP_TEMPLATE"
+	ArtifactCopyStepCABundleConfigMapName   string = "ARTIFACT_COPY_STEP_CABUNDLE_CONFIGMAP_NAME"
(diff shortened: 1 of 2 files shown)


Tomcli (approved):
/lgtm
/approve

Thanks @HumairAK

PR #1426, opened 2023-12-06
feat(tekton-kfptask): Update kfptask to publish completed dag status

Tomcli:
Resolves #

Our current publisher only being executed when the dag is completed. Thus it make sense to also update mlmd status as complete for that dag. 

For failing status, caching entry, and dag output artifacts, it only make sense to implement it after we merged the publisher as part of the sub-dag reconciliation logic. Otherwise any failed sub-dag will never run the publisher task and the cache entry and output artifacts info has different logic when it became part of the sub-dag reconciliation logic.

**Environment tested:**

**Checklist:**


Tomcli:
/cc @yhwang 
This is just to show we can update the dag publisher when it's completed. It cannot be a complete implementation using the current publisher design as the failed dag never able to run the publisher task and update the dag.

This is a POC design and I will try to propose something more complete to the KFP community's driver package when we have a more completed design.

yhwang (approved):
/lgtm

PR #1455, opened 2024-02-12
fix(README): Update instructions to only use kustomize

Tomcli:
Resolves #

Updating install instructions to use kubectl -k. kustomize is now a de facto part of kubectl and it make sense to let users install the manifests from it's true source rather than the release yaml we created. That way when master branch is cleaned up, users can still get the source file from each release tag.

**Environment tested:**

**Checklist:**


yhwang:
this change reminds me that we don't need to run those GH actions if the content is doc/readme change only.

